PR #265, opened 2015-02-16
Issue/259 android

ichinaski:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

app/src/main/java/org/akvo/flow/api/FlowApi.java
```diff
@@ -161,5 +150,8 @@ private String getTimestamp() {
         String TIMESTAMP    = "ts=";
         String LAST_UPDATED = "lastUpdateTime=";
         String HMAC         = "h=";
+        String VERSION      = "ver=";
+        String DEVICE_ID    = "devId=";
+        String ANDROID_ID   = "androidId=";
```
muloem: The corresponding changes for these new parameters are going into the servlet right?

app/src/main/java/org/akvo/flow/service/ExceptionReportingService.java
```diff
@@ -153,6 +153,8 @@ public void submitStackTraces(String server) {
                     File f = new File(dir, list[i]);
                     String trace = FileUtil.readFileAsString(f);
 
+                    // We cannot use the standard FlowApi.getDeviceParams, fot this service uses
+                    // a different naming convention...
                     Map<String, String> params = new HashMap<String, String>();
```
muloem: So we don't introduce the `ANDROID_ID` here because of the convention or because its not needed?

app/src/main/java/org/akvo/flow/api/FlowApi.java
ichinaski: `ver` and `devId` are already being read by the servlet (I just refactored the definition). But indeed, `androidId` is the new parameter that will need to be parsed. Might be wise to document this somewhere.

app/src/main/java/org/akvo/flow/service/ExceptionReportingService.java
ichinaski: Apparently this servlet uses it's own convention, thus it is probably better off without any extra complexity. However, it might be a good idea to send the `androidId` param, even if its not read.

muloem:
:+1: code reviewed!

PR #344, opened 2015-09-01
Feature/257 repeat question groups 2

ichinaski:


mtwestra:
@ichinaski we are testing the app at uat1 at the moment. This one still creates repeats with the iteration number starting at -1. Do you perhaps know why?

PR #427, opened 2016-03-11
Hotfix/2.2.3.1

ichinaski:
### Redefine response validity check

In order to consider the response as valid, we check:
* Generic response is valid (`super.isValid()`). This will make sure mandatory questions have a response
* Response is complete (`mFinished`). Partial responses will invalidate the response

Additionally, on line 126 we set the completeness flag to `false`, for the current level has no selection.


muloem:
:+1: code reviewed!

Issue #433, opened 2016-05-02
Disable barcode response manual edits


ichinaski:
Leverage existing `geoLocked` property, currently used on GeoQuestion and GeoShapes

ichinaski:
**Test plan:**
* Create a barcode question and disable manual input
* Field should not be focusable

rumca:
test plan passes as described

PR #434, opened 2016-05-06
[#433] Disable barcodes when question is locked

ichinaski:


stellanl:
Code checked.
Copyright date is not updated.

Issue #454, opened 2016-07-12
Do not check for new surveys before Device ID is set


stellanl:
The cause of akvo/akvo-flow#1731
When app is run for the first time, it puts up AddUserActivity to ask for user name and device id, but in the background SurveyDownloadService immediately queries the server for surveys, at which point it appears in the device list. Unless there are network problems, this happens before the device ID has been set, so until the next survey check, it will be in the device list without the user-set ID. AddUserActivity does not, in contrast with PreferencesActivity, start SurveyDownloadService to propagate the new device ID.

Suggested solution:
1. In SurveyActivity, if AddUserActivity is called, skip starting SurveyDownloadService.
2. In AddUserActivity, start SurveyDownloadService when a device ID has been set. (Exactly like in PreferencesActivity.)
Alternatively, we could do just 2. which will at least minimise the window during which a device without an ID is visible.

janagombitova:
No test plan, but the device ID is shown right away in the device list with the device as well. So passes 👍

PR #455, opened 2016-07-18
[#454]On first run, start SurveyDownloadService after DeviceID is set

stellanl:


app/src/main/java/org/akvo/flow/activity/SurveyActivity.java
```diff
@@ -127,12 +127,14 @@ public void onDrawerOpened(View drawerView) {
...
         // Start the setup Activity if necessary.
+        boolean noDevIdYet = false;
         if (!Prefs.getBoolean(this, Prefs.KEY_SETUP, false)) {
+            noDevIdYet = true;
             startActivityForResult(new Intent(this, AddUserActivity.class), REQUEST_ADD_USER);
         }
 
         displaySelectedUser();
-        startServices();
+        startServices(noDevIdYet);
```
mtwestra: Isn't noDevIdYet still true here? In that case, the service will not run?

stellanl: Yes, that is the idea. It will not be started here, but only on return from the get-user-and-devid screen.

app/src/main/java/org/akvo/flow/activity/AddUserActivity.java
```diff
@@ -61,6 +63,10 @@ public void onClick(View v) {
                 db.savePreference(ConstantUtil.DEVICE_IDENT_KEY, deviceId);
                 db.close();
 
+                // Trigger the SurveyDownload Service, in order to make
+                // a backend connection with the new Device ID
```
mtwestra: Isn't it strange to trigger the surveydownload service within the adduseractivity? It seems to mix things that are unrelated?

stellanl: I guess it could be moved to where SurveyActivity receives the return value from AddUserActivity. I was sticking as close as possible to the similar code in PreferencesActivity.

mtwestra: I guess I would be in favour of moving it there, as it will clarify the intent and structure of the code.

stellanl:
Should be merged into develop, not master

PR #465, opened 2016-08-17
[#454] Wait for Device ID setting before connecting to dashboard (connect #454)

stellanl:
Now based on develop


app/src/main/java/org/akvo/flow/activity/AddUserActivity.java
```diff
@@ -16,6 +16,7 @@
 
 package org.akvo.flow.activity;
 
+import android.content.Intent;
```
ichinaski: Unused import?

app/src/main/java/org/akvo/flow/activity/AddUserActivity.java
```diff
@@ -29,6 +30,7 @@
 import org.akvo.flow.app.FlowApp;
 import org.akvo.flow.dao.SurveyDbAdapter;
 import org.akvo.flow.domain.User;
+import org.akvo.flow.service.SurveyDownloadService;
```
ichinaski: ditto

Issue #487, opened 2016-11-07
When error synchronizing a file, the error message is hardcoded and not translated


valllllll2000:
We should extract that string to resources and translate it:
"Form " + formId + " does not exist", "It has probably been deleted");

Test plan:
This error is shown when the form has been deleted in the dashboard, thus we cannot sync it.

janagombitova:
@valllllll2000 When trying to download a non-existent form (I created one on the dashboard, wrote down the form ID, deleted the full survey, used manual download on the app to download the form) I received an this error message: Form xyz could not be read from F.. > unfortunately I cannot see the full text (another issue to tackle) 

![screenshot_2016-11-21-12-15-56](https://cloud.githubusercontent.com/assets/12456965/20481007/379384ba-afe5-11e6-97d6-e1091210df32.png)

Are you unifying the notification text with this issue as well? If so, it is different from the example above and from the strings that are currently in Transifex.

valllllll2000:
@janagombitova This issue is only about the text itself. I have created a separate issue for the single line issue : https://github.com/akvo/akvo-flow-mobile/issues/519

also this string you are showing in the screenshot is a different case. I have not touched that one. 

What has been done:
The 2 strings they already existed in the app but they were not used in the correct way for Android and due to that they had no translation. I have only extracted the string, I have not modified functionality or the string itself. From what is written in the code is:
This string is when we try to sync forms, here is a description of the class:
 ```
Handle survey export and sync (...). The export process  (...) will try to zip all the survey instances with a SUBMITTED status but with no EXPORT_DATE (export hasn't happened yet). (...) if for
 whatever reason, a previous export attempt has failed, a new export will be tried (...) until the zip file finally gets exported. (...)

 After the export of the zip files, the sync will be run, attempting to upload all the non synced
 files to the datastore.
```
I have remove the bits which are too technical. It seems to me this tries to upload non synced forms but the message is shown because the survey has been deleted?

janagombitova:
@valllllll2000 thank you for the explanation and for creating the issue (I Have referenced and old one to it). I misunderstood the steps to verify. Following your explanation all works as expected 👍 

Interestingly, the users gets a notification about the deletion of a form, but he can still go and submit the form, which then does not get synchronised. It would be good to save him the trouble of even trying to fill in the form, if he cannot sync it in the 1st place (by somehow not allowing them to open up the form at all) Will create an issue for this in product design as we need more UX input here.

PR #488, opened 2016-11-07
Extracted error string to resources (connect #487)

valllllll2000:


valllllll2000:
@muloem I have moved the test plan and updated the year.

Issue #486, opened 2016-11-07
Error notification icon is shown as empty white circle


valllllll2000:
## The issue:
Turns out as the icon id is passed as null for the notification when there was an error downloading forms.

When this id is null, R.drawable.info is used, but it is only defined for drawable-hdpi.

After talking to @janagombitova it has been decided we should use the akvo icon. 

## What has been done:
Modify the way notifications are shown.
Added new akvo icon.
Informative/progress notifications shown in orange.
Error notifications are in red.
There is a difference between how they are shown in different android versions (see screenshots below).
Older devices will only show a white icon (no color available)

## Test plan:
Ideally test all case where the notifications are displayed, either success or error.
If possible test on older devices (< android 5) and also android 5/6 and 7.

valllllll2000:
![screenshot_20161107-141030](https://cloud.githubusercontent.com/assets/923280/20059237/698e2b28-a4f5-11e6-9433-2f48a0fdd3e6.png)

valllllll2000:
@janagombitova after looking at the other icons used (see for example the one on the screenshot for Downloading forms, we should use a white icon, the akvo app icon cannot be used here. The proposed icon to use is android.R.drawable.ic_dialog_alert, see my screenshot below of how it will look:

![screenshot_20161107-144112](https://cloud.githubusercontent.com/assets/923280/20059916/ebc8d23e-a4f8-11e6-831a-97a7b679db01.png)

valllllll2000:
after talking to @loicsans it has been decided to use the same icon for all notification icons:

![imgpsh_fullsize](https://cloud.githubusercontent.com/assets/923280/20063653/4b899ef6-a508-11e6-91a6-4ce287ecd4d5.png)

muloem:
@loicsans Shouldn't we distinguish between the kinds of notifications based on what the error actually is?  if it is simply informational notification, I agree that we can use the Akvo icon but I think error notifications could perhaps stand out more.

Also is it possible for the icon to be coloured or does that look weird at that small size?

@valllllll2000 @janagombitova

valllllll2000:
@muloem the icon has to be black and white as it is used for the status bar. A colored icon just shows an empty white square.

muloem:
ok.. and having an orange background?

valllllll2000:
I can try @muloem, this usually works on lollipop and on Nougat the icon can be colored as well but no background, I will try now and upload some screenshots to see how it looks.

[...]

PR #489, opened 2016-11-07
Error notification icons (connect #486)

valllllll2000:
Modify the way notifications are shown. 
Added new icon
Akvo notifications shown in orange
Error notifications are in red
Older devices will only show a white icon (no color available)


valllllll2000:
@muloem I have made the changes you requested. I also have uploaded some screenshots to the issue so if they are accepted I will push the changes.

valllllll2000:
@muloem I have made the changes after Loic's comment on the issue so now here is how both normal and error notifications look. You can now take a look at the code again.
notifications

app/src/main/java/org/akvo/flow/util/NotificationHelper.java
```diff
@@ -0,0 +1,147 @@
+package org.akvo.flow.util;
+
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.content.Intent;
+import android.support.v4.app.NotificationCompat;
+import org.akvo.flow.R;
+
+/*
```
valllllll2000: Fixed!

valllllll2000:
Fixed @muloem

PR #495, opened 2016-11-10
Create PULL_REQUEST_TEMPLATE.md

valllllll2000:
The objective of this is to have always the same format on the PR and when someone presses create pull request, this data will be pre-filled.

What do you guys think, is there a missing info? Something should be rephrased?

Do you think it will be useful?

If so this may be extended to other Akvo repos.

More info on github templates:
https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,5 @@
 #### Before the PR (what is the issue or what needed to be done)
 #### The solution
-#### Test plan
 #### Screenshots (if appropriate)
 
 Remember to connect the issue #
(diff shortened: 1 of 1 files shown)


stellanl:
For the test plan we already have akvobot nagging us about it. So far I have always put test plans into comments of the issue, not the pull requests.

janagombitova:
For Documentation, I am not sure which documentation it refers to, if it is support documentation for users then I take care of it and I also do it after testing the issue, looking at the test plan in the issue

valllllll2000:
@stellanl @janagombitova so I have updated the PR to remove the Test plan from the PR template. I still think a template is useful but feel free to comment / accept or even refuse.

Issue #499, opened 2016-11-10
When device is rotated user is shown the download update dialog again


valllllll2000:
Steps to reproduce:
Install an app for which there should be an update
Open app
you will see the dialog of the screenshot
Press no
Rotate device
You will see the dialog again 

This could be annoying of the user does not want or cannot update now so we have decided with @janagombitova that we will display the dialog only once every 24 hours if the user pressed "no".

Of course for the user to be able to update later we need to add a way to update from the settings screen (see issue https://github.com/akvo/akvo-flow-mobile/issues/500)

See screenshot:
![screenshot_20161110-162227](https://cloud.githubusercontent.com/assets/923280/20183380/60f326e0-a765-11e6-9c5a-7dde8a61d619.png)

valllllll2000:
Test plan:
Open the app that has not yet been updated (we will probably need a special build for that), the dialog will be shown, press no, rotate the device, make sure the dialog is not shown a second time. Come back after 24h and open the app. The dialog should be shown.

janagombitova:
@valllllll2000 is this not handled with the implementation in the other issues? or in the apk that I tested?

valllllll2000:
@janagombitova this is not finished. I made some changes but there is a class present in newer google api which would be perfect for this case so I am waiting on some decision to be taken on that other task.

valllllll2000:
This ha been developed as discussed previously with @janagombitova but after talking to @muloem he has suggested we should maybe not block the user at all but display a notification and that maybe a check once a week will be sufficient. In the meantime this is still an improvement over what is currently there.

janagombitova:
I can confirm that nothing extra is shown to the user when the device is rotated. Will cross check this issue once more after I have the 2.2.9 build to be able to update the version. Leaving the issue in `ready for testing` for the time being

janagombitova:
👍  all ready to go

PR #506, opened 2016-11-15
Show update dialog only once (connect #499)

valllllll2000:
When changing orientation each time the user gets shown a dialog to update
#### The solution
The dialog will be shown only once every 24h. 
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


valllllll2000:
Closing for now as not yet satisfied with the solution

app/src/main/java/org/akvo/flow/activity/AppUpdateActivity.java
```diff
@@ -175,26 +206,27 @@ protected void onProgressUpdate(Integer... progress) {
...
         }
 
         @Override
         protected void onPostExecute(String filename) {
-            if (TextUtils.isEmpty(filename)) {
-                Toast.makeText(AppUpdateActivity.this, R.string.apk_upgrade_error, Toast.LENGTH_SHORT).show();
-                mInstallBtn.setText(R.string.retry);
-                mInstallBtn.setEnabled(true);
-                return;
+            if (activityWeakReference.get() != null) {
```
stellanl: I'm no expert in weak references, but isn't it theoretically possible that the activity could be garbage collected between these two calls to get() ? To close this loophole you need to assign the result of the (first) get() to a variable, which you then test **and** use.

app/src/main/java/org/akvo/flow/activity/AppUpdateActivity.java
```diff
@@ -226,13 +255,10 @@ private File createFile(String location, String version) {
...
          * After the download, a new notification will be displayed, requesting
          * the user to 'click to installAppUpdate'
-         *
-         * @param remoteFile
-         * @param surveyId
          */
         private boolean downloadApk(String location, String localPath) {
             Log.i(TAG, "App Update: Downloading new version " + mVersion + " from " + mUrl);
-            if (!StatusUtil.hasDataConnection(AppUpdateActivity.this)) {
+            if (activityWeakReference.get() != null && !StatusUtil.hasDataConnection(activityWeakReference.get())) {
```
stellanl: Same problem here - each call to get() could be the first to return a null, meaning that the Activity was just collected.

Issue #526, opened 2016-11-23
Syncing imported data points


janagombitova:
For monitored surveys the app user can sync data points collected by other users. This functionality works fine. However, data points can also be created via the Data cleaning functionality. The app user can also sync these, the same way he would sync points collected via the app. He should not see the difference on the device. 

## Example case:
- uat1.akvoflow.org > 1.9.12 UAT folder > 1.9.12 UAT test > 1.9.11 MF Survey copy > registration form 
   - 6 data points uploaded via data cleaning 
- uat1.akvoflow.org > 1.9.12 UAT folder > 1.9.12 UAT test > 1.9.11 All questions survey copy > All questions form 
   - 3 data points uploaded via data cleaning 

All points work fine on the dashboard and Flow services 

## Steps take on the app
1. After uploading the data points I selected the survey on the app
2. Clicked on Sync
3. Data points synced is successfully based on notification message
4. Data points appear in the list of points but have the icon "saved" 

![screenshot_2016-11-23-17-11-31](https://cloud.githubusercontent.com/assets/12456965/20569694/428ab0ce-b1a1-11e6-9d77-9638082467b9.png)

5. If I click on one of the points, it seems as no forms are submitted 

![screen shot 2016-11-23 at 17 22 21](https://cloud.githubusercontent.com/assets/12456965/20569747/7976e9b8-b1a1-11e6-8cf1-0e9bf25cca57.png)

6. After a few seconds the points disappear from my list on the phone

janagombitova:
After a few more attempts with @valllllll2000 we learned that the issue happens, if the user syncs right after the upload is finalised. In that case the issue occurs. If he waits a minute longer, the sync is completed without problems

valllllll2000:
To provide more info here is what we get from the api in this case:
`{"surveyedLocaleData":[{"id":"e57u-g3hq-m79d","displayName":"","identifier":null,"lat":null,"lon":null,"surveyInstances":[],"surveyGroupId":98729115,"lastUpdateDateTime":1479985176732,"keyId":null}],"message":null,"offset":0,"code":"200","resultCount":1}`
As you can see there are no actual answers to that survey but there should be.

valllllll2000:
With @janagombitova we have decided that in the app what can be done is to not display a datapoint which does not have any answers and display an error message for the user to attempt sincing again.

valllllll2000:
This is the message that would be displayed:
Some items were not synced correctly. Please sync again.

[...]

PR #528, opened 2016-11-25
Syncing imported datapoints (connect #526)

valllllll2000:
Data points which were imported using the data cleaning tool if synced right after on device turned out to be corrupted. They were shown as saved (see screenshot in the issue)
#### The solution
Cleanup database after syncing locales to remove datapoints with no responses.
Changed the status to start with 1 since 0 is usually the default in the db to avoid confusion.
Notify user that sync failed so he can sync again and hopefully get the correct data.
Some cleanup and small code refactor.
NA
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/api/FlowApi.java
```diff
@@ -16,67 +16,198 @@
...
+    public List<Survey> getSurveyHeader(String serverBaseUrl, String surveyId) throws IOException {
+        final String url = serverBaseUrl + SURVEY_HEADER_SERVICE_PATH + surveyId + "&" + getDeviceParams();
+        String response = HttpUtil.httpGet(url);
+        if (response != null) {
+            return new SurveyMetaParser().parseList(response, true);
+        }
+        return Collections.emptyList();
+    }
+
+    public List<Survey> getSurveys(String serverBase, List<Survey> surveys) throws IOException {
```
muloem: I was wondering whether here it's ok to eliminate the second parameter since we will return the list of surveys anyway. i.e we done modify it but rather overwrite the reference.

app/src/main/java/org/akvo/flow/dao/SurveyDbAdapter.java
```diff
@@ -171,11 +176,12 @@
     }
 
     public interface SurveyInstanceStatus {
-        int SAVED      = 0;
-        int SUBMITTED  = 1;
-        int EXPORTED   = 2;
-        int SYNCED     = 3;
-        int DOWNLOADED = 4;
+        int UNKNOWN = 0;
```
muloem: How will changing these affect any data that is already in the app when the update happens?

muloem:
Oh and there are some merge conflicts to sort out of course

app/src/main/java/org/akvo/flow/api/FlowApi.java
```diff
@@ -16,67 +16,198 @@
...
+    private static final String SURVEY_HEADER_SERVICE_PATH = "/surveymanager?action=getSurveyHeader&surveyId=";
+    private static final String NOTIFICATION_PATH = "/processor?action=";
+    private static final String DEVICE_NOTIFICATION_PATH = "/devicenotification";
+    // Sync constants
+    private static final String FILENAME_PARAM = "&fileName=";
+    private static final String FORMID_PARAM = "&formID=";
+    private static final int ERROR_UNKNOWN = -1;
+    private static final String HMAC_SHA_1_ALGORITHM = "HmacSHA1";
+    private static final String BEACON_SERVICE_PATH = "/locationBeacon?action=beacon";
+    private static final String LAT = "&lat=";
```
valllllll2000: You are totally right I have made the changes

app/src/main/java/org/akvo/flow/api/FlowApi.java
```diff
@@ -16,67 +16,198 @@
...
+
+    /**
+     * Request the notifications GAE has ready for us, like the list of missing files.
+     *
+     * @return String body of the HTTP response
+     *
+     * @throws Exception
+     */
+    @Nullable
+    public JSONObject getDeviceNotification(@NonNull String serverBase, @NonNull String surveyIds) throws Exception {
```
valllllll2000: Good, point I have changed it

app/src/main/java/org/akvo/flow/api/FlowApi.java
valllllll2000: Done

app/src/main/java/org/akvo/flow/dao/SurveyDbAdapter.java
valllllll2000: I got scared and I have reverted back since I am not sure 100% that it may not break something that we will not see immediately. I have in project to refactor the database class and then I will test it fully.

valllllll2000:
@muloem all fixed!

Issue #521, opened 2016-11-22
Form name does not wrap


joycarpediem:
![screenshot_2016-11-22-08-13-16](https://cloud.githubusercontent.com/assets/4419787/20507223/e8bc5478-b08b-11e6-9786-766d280fd8b7.jpg)

valllllll2000:
@joycarpediem Thank you for reporting this. Could you please tell me on which device and android version you found this?

joycarpediem:
@valllllll2000 Android 5.1.1 Language English(UK)

janagombitova:
Wraps well, in both menu and also list of forms. 👍 passes test plan perfectly  

![screen shot 2017-01-02 at 15 39 23](https://cloud.githubusercontent.com/assets/12456965/21591143/afd09598-d101-11e6-9a9c-2321cb8f9a0a.png)

PR #566, opened 2016-12-14
Form name wrap issue (connect #521)

valllllll2000:
Form name was not bieng wrapped in some cases due to the fact that 2 diferent textviews were being used and if the first one was too large, the second one was displayed in a strange way (see screenshots on the issue)
#### The solution
Use spannable and only one TextView to prevent that issue. Spannable allows to have diferent text sizes within one textView for example.
screenshot from 2016-12-14 18-45-09

#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/app/src/main/res/layout/survey_item.xml
+++ b/app/src/main/res/layout/survey_item.xml
@@ -1,33 +1,24 @@
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:minHeight="?android:attr/listPreferredItemHeight"
-    android:gravity="center_vertical"
-    android:padding="8dp" >
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                xmlns:tools="http://schemas.android.com/tools"
+                android:minHeight="?android:attr/listPreferredItemHeight"
+                android:gravity="center_vertical"
+                android:padding="8dp">
 
     <TextView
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/survey_item.xml
```diff
@@ -1,33 +1,24 @@
...
-        android:layout_height="wrap_content"
-        android:layout_marginLeft="8dp"
-        android:layout_marginStart="8dp"
-        android:layout_toRightOf="@id/text1"
-        android:layout_toEndOf="@id/text1"
-        android:layout_alignBaseline="@+id/text1"
-        android:enabled="false"
-        android:textSize="18sp" />
+        android:textColor="@color/text_color_primary"
+        tools:text="Farmer Rice Diary: Pest - Disease"
```
muloem: Is this a placeholder or just something accidentally left over?

muloem:
@valllllll2000 beside the comment I left, I see no other issues.

app/src/main/res/layout/survey_item.xml
valllllll2000: @muloem tools prefix allows you to display a text only in the xml editor (layout preview), the users will never see it. more info here: https://developer.android.com/studio/write/tool-attributes.html check the tools instead of android section.

muloem: ahh! got it!

Issue #592, opened 2017-01-12
Update repo license to GPL


muloem:
We need to change the license of the mobile app code to GPL and include the correct license files and headers.

valllllll2000:
@muloem Do you have a link as an example where I get them? Or is it just replacing the licence name?

Here is the current text:
`/*
 * Copyright (C) 2010-2017 Stichting Akvo (Akvo Foundation)
 *
 * This file is part of Akvo FLOW.
 *
 * Akvo FLOW is free software: you can redistribute it and modify it under the terms of
 * the GNU Affero General Public License (AGPL) as published by the Free Software Foundation, either version 3 of the License or any later version.
 *
 * Akvo FLOW is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Affero General Public License included below for more details.
 *
 * The full license text can also be seen at <http://www.gnu.org/licenses/agpl.html>.
 *
 */`

Shall I replace AGPL by GPL and that is it?

muloem:
So for now, 

* [ ] we need to replace two things _FLOW_ with _Flow_, and AGPL with GPL
* [ ] remove the "Affero"
* [ ] replace the AGPL link with the link to GPL

muloem:
>     This file is part of Akvo Flow.
> 
>     Akvo Flow is free software: you can redistribute it and/or modify
>     it under the terms of the GNU General Public License as published by
>     the Free Software Foundation, either version 3 of the License, or
>     (at your option) any later version.
> 
>     Akvo Flow is distributed in the hope that it will be useful,
>     but WITHOUT ANY WARRANTY; without even the implied warranty of
>     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
>     GNU General Public License for more details.
> 
>     You should have received a copy of the GNU General Public License
>     along with Akvo Flow.  If not, see <http://www.gnu.org/licenses/>.
>

stellanl:
So:
* [ ] Drop in the GPL file in the source root.
* [ ] Change the copyright headers of all the Java files (and any other source)

janagombitova:
@stellanl is there something I need to test here?

valllllll2000:
@janagombitova this is not included in the release build

stellanl:
Not really, none of it is visible in the product, just in the repo...

PR #594, opened 2017-01-12
Issue/592 agpl to gpl (Connect #592)

stellanl:
Most files had AGPL headers, a few had none or headers unchanged from Eclipse templates.
#### The solution
Changed all headers to GPL3, changed name in copyright notice from FLOW to Flow, added headers where missing, made formatting more consistent.
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Copyright header
* [ ] Code formatting

### DIFF
--- a/app/src/main/java/org/akvo/flow/activity/AddUserActivity.java
+++ b/app/src/main/java/org/akvo/flow/activity/AddUserActivity.java
@@ -1,17 +1,20 @@
 /*
  *  Copyright (C) 2015-2016 Stichting Akvo (Akvo Foundation)
  *
- *  This file is part of Akvo FLOW.
+ *  This file is part of Akvo Flow.
  *
- *  Akvo FLOW is free software: you can redistribute it and modify it under the terms of
- *  the GNU Affero General Public License (AGPL) as published by the Free Software Foundation,
- *  either version 3 of the License or any later version.
+ *  Akvo Flow is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
  *
- *  Akvo FLOW is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
- *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
- *  See the GNU Affero General Public License included below for more details.
+ *  Akvo Flow is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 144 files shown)


app/src/main/java/org/akvo/flow/activity/AddUserActivity.java
```diff
@@ -1,17 +1,20 @@
 /*
  *  Copyright (C) 2015-2016 Stichting Akvo (Akvo Foundation)
```
valllllll2000: Should be 2017

app/src/main/java/org/akvo/flow/activity/AppUpdateActivity.java
```diff
@@ -1,17 +1,20 @@
 /*
  *  Copyright (C) 2014-2016 Stichting Akvo (Akvo Foundation)
```
valllllll2000: Should be 2017

app/src/main/java/org/akvo/flow/activity/BackActivity.java
```diff
@@ -1,18 +1,20 @@
 /*
- * Copyright (C) 2010-2016 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2016 Stichting Akvo (Akvo Foundation)
```
valllllll2000: Should be 2017

app/src/main/java/org/akvo/flow/activity/GeoshapeActivity.java
```diff
@@ -1,18 +1,22 @@
 /*
  *  Copyright (C) 2015 Stichting Akvo (Akvo Foundation)
```
valllllll2000: update the year, should it be 2015-2017?

muloem:
@valllllll2000 @stellanl FYI the copyright year should only be updated in case that source code in that file was actually modified that year. so if there are files that were modified in 2017 update them otherwise the Copyright year (range) should remain as is.

valllllll2000:
@muloem of course but in this case, he modified the copyright header of all the files so the files were modified so it should be updated, right? Or does the copyright notice not count as modification?

muloem:
Given that we are copyrighting IP, I wouldn't count editing the copyright header as a modification.

valllllll2000:
@stellanl @muloem ok I will remove my comments then and review the PR again

app/src/main/java/org/akvo/flow/api/S3Api.java
```diff
@@ -1,3 +1,22 @@
+/*
+ *  Copyright (C) 2013 Stichting Akvo (Akvo Foundation)
```
valllllll2000: Why does it say 2013 here? If you just added this copyright notice, should it be 2017? or 2010-2017

stellanl:
I did not touch the copyright year in the headers at all

Issue #597, opened 2017-01-16
Change colour in geoshape editor


janagombitova:
I believe this issue relates to the other changes in the apps UI we experienced with the last updates. In the test 2.2.10 apk the geoshape edit icons look grey and seem to be disabled even thought that is not the case. 

![screenshot_2017-01-16-14-45-28](https://cloud.githubusercontent.com/assets/12456965/21985386/aabce3fa-dbfa-11e6-8214-09f40838825b.png)

Before the change the icons in the bottom of the screen were better visible. Can you please change this? 

![screenshot_2017-01-16-15-40-04](https://cloud.githubusercontent.com/assets/12456965/21987037/2adb3a8a-dc02-11e6-80d3-64d3da6bda08.png)

valllllll2000:
Test plan:
* Create a survey which requires geoshapes
* Open it on your device and go to "capture shape"
* Chose the menu option for shapes and chose between Line/Points or Area

janagombitova:
👍  passes test plan, ready to go

PR #598, opened 2017-01-16
Fixing the geoshape image button styling (connect #597)

valllllll2000:
The geoshape bottom image buttons started to look disabled after support library update
#### The solution
created new background and styling for the image buttons
In the screenshot, you can see that the 3rd button is disabled which is normal.
geoshapeimagebuttons

#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


janagombitova:
This it how it was before as well. I agree with you @stellanl this needs a better design to it and we can create a separate issue to address this, but for now the goal is to bring the icons back to their original state for this release

Issue #614, opened 2017-01-31
Offline survey export: files folder missing


valllllll2000:
From remaze request # 41,280.
Filled surveys should be located in the folder akvoflow/data/files for offline transfer to PC.

With the version 2.2.10 this folder is missing.

There is no error displayed instead the notification shows "data exported to sd card...".
Users are unable to find the .zip file with the filled survey.
Reproduced on Samsung J7 with android 6.0 and Nexus 6 with android 7.1.1. 

Steps to reproduce:
Delete the akvoflow folder
Install the app
Make sure you have at least a survey assigned.
Put device in flight mode.
Fill the survey and press "send".
Connect device via usb to PC. You will see that the folder akvoflow/data/files is missing.

Update: It seems this issue is not about what we originally expected.
After connecting to the partners device, the folder akvoflow/data/files is actually mising. The transmission history is empty:
![screenshot from 2017-02-02 13-42-42](https://cloud.githubusercontent.com/assets/923280/22586514/7691f708-e9fd-11e6-813b-6ffb48930a13.png)
The surveys seem in an exported state:
![screenshot from 2017-02-02 13-41-24](https://cloud.githubusercontent.com/assets/923280/22586546/a90c2352-e9fd-11e6-8f4d-6636938980fe.png)
User is able to view the collected surveys on the device.
In total 11 devices have the issue of missing folder:
[phone issues.xlsx](https://github.com/akvo/akvo-flow-mobile/files/750038/phone.issues.xlsx)

valllllll2000:
Update: there is no error logs and connecting to the device via usb the files are there:
angler:/sdcard/akvoflow/data $ ls
files media 
angler:/sdcard/akvoflow/data $ cd files/                                       
angler:/sdcard/akvoflow/data/files $ ls
02c3ff8a-b36f-4192-b239-5128f6217bb2.zip 
30a7149d-07cd-4431-a5dc-f3a608630fb0.zip 
58f99c58-ec5e-45a3-8386-90e012bba185.zip 
5eb9ed91-064d-48ba-836f-68bfcb58fbee.zip 
9db4743d-fba6-4745-83e6-9196e2feb226.zip 
d6eba5d7-e092-43fa-8304-faa2ce9fb1fc.zip 

They just do not seem to be visible when browsing the device filesystem but it does work if I try to actively pull the .zip file from device, for example:
[...]

PR #619, opened 2017-02-07
Fix issue with strange bootstrap folders (connect #614)

valllllll2000:
When using bootstrap files, in some cases the folder structure is surveyId/survey_name/survey_name.xml
In that case the survey_name was being used as surveyId which broke the creation of filled surveys zip files
#### The solution
1) Make sure that surveyId is a String (and not a Long)
2) Do not stop file export if one of the surveys fails, continue with next one
3) Fix the weird folder not an id and getting the id from the correct folder and also being able to get survyeId from xml file when present.
NA
#### Reviewer Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/java/org/akvo/flow/service/BootstrapService.java
+++ b/app/src/main/java/org/akvo/flow/service/BootstrapService.java
@@ -236,9 +236,9 @@ public class BootstrapService extends IntentService {
     }
 
     private void processSurveyFile(@NonNull ZipFile zipFile, @NonNull ZipEntry entry,
-            @NonNull String filename, @NonNull String id) throws IOException {
+            @NonNull String filename, @NonNull String idFromFolderName) throws IOException {
 
-        Survey survey = databaseAdapter.getSurvey(id);
+        Survey survey = databaseAdapter.getSurvey(idFromFolderName);
 
         String surveyFolderName = generateSurveyFolder(entry);
 
@@ -254,7 +254,7 @@ public class BootstrapService extends IntentService {
 
         verifyAppId(basicSurveyData);
 
-        survey = updateSurvey(filename, id, survey, surveyFolderName, basicSurveyData);
+        survey = updateSurvey(filename, idFromFolderName, survey, surveyFolderName, basicSurveyData);
 
         // Save the Survey, SurveyGroup, and languages.
         updateSurveyStorage(survey);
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/akvo/flow/util/SurveyIdGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Get the survey id from the survey xml folder
+     * The structure can be either surveyId/survey.xml or surveyId/folder/survey.xml
+     *
+     * @param parts
+     * @return
+     */
+    @NonNull
+    public String getSurveyIdFromFilePath(@Nullable String[] parts) {
+        if (parts == null || parts.length <= 1) {
```
valllllll2000: For now, parts.lenght == 1 there is only the file name but when I make the change you requested, then this will also change.

app/src/main/java/org/akvo/flow/service/BootstrapService.java
```diff
@@ -171,7 +176,7 @@ private void processFile(File file) throws Exception {
             Log.d(TAG, "Processing entry: " + entry.getName());
             String parts[] = entry.getName().split("/");
             String filename = parts[parts.length - 1];
-            String id = parts.length > 1 ? parts[parts.length - 2] : "";
+            String id = surveyIdGenerator.getSurveyIdFromFilePath(parts);
```
valllllll2000: Good idea, I have simplified my logic as you suggested.

valllllll2000:
@muloem all fixed now, thanks.

app/src/main/java/org/akvo/flow/util/SurveyIdGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ */
+
+package org.akvo.flow.util;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.text.TextUtils;
+
+public class SurveyIdGenerator {
```
muloem: Based on the changes, I am not sure we need this class

muloem: Scratch this comment..

PR #623, opened 2017-02-14
More reliable bootstrap files (connect #614)

valllllll2000:
Some fixes were needed to the logic since the merge of this branch
#### The solution
filename and id only generated for xml and help files
filename generation moved to a delegate + tests for this delegate
NA
#### Reviewer Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/util/SurveyFileNameGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+import android.support.annotation.Nullable;
+
+public class SurveyFileNameGenerator {
+
+    public SurveyFileNameGenerator() {
+    }
+
+    @NonNull
+    public String generateFileName(@Nullable String entryName) {
+        if (entryName == null || entryName.isEmpty()) {
```
muloem: I would suggest we add the conditions of this block at the beginning of the `processFile()` method where we have other checks e.g. `entry.isDirectory()` otherwise it would be undesirable behaviour to have a survey in the table with an empty string as the name.

And in that case specify `entryName` as non-nullable?

app/src/main/java/org/akvo/flow/util/SurveyIdGenerator.java
```diff
@@ -33,12 +33,15 @@ public SurveyIdGenerator() {
...
      *
-     * @param folderPath all the folder path expluding the actial fileName
+     * @param entryName all the folder path expluding the actial fileName
      * @return
      */
     @NonNull
-    public String getSurveyIdFromFilePath(@Nullable String folderPath) {
-        String folders[] = folderPath == null ? null : folderPath.split("/");
+    public String getSurveyIdFromFilePath(@Nullable String entryName) {
+        int fileSeparatorPosition = entryName == null? -1 : entryName.lastIndexOf("/");
```
muloem: Ditto here re `indexOf()`

And if the setup is always `surveyId/folder`, then I would pick everything up to the first "/" and take that as the id. And if we can't figure out the id, return a `null`

app/src/test/java/org/akvo/flow/util/SurveyIdGeneratorTest.java
```diff
@@ -79,28 +79,32 @@ public void getSurveyIdFromFilePath_shouldReturnCorrectIdFromFolder() throws Exc
...
 
         String surveyId = surveyIdGenerator
-                .getSurveyIdFromFilePath("123");
+                .getSurveyIdFromFilePath("123/file.xml");
         assertEquals("123", surveyId);
 
         surveyId = surveyIdGenerator
-                .getSurveyIdFromFilePath("folder/123");
+                .getSurveyIdFromFilePath("folder/123/file.xml");
         assertEquals("123", surveyId);
```
muloem: Is this scenario possible? i.e one where the surveyId folder is a subfolder?

valllllll2000: Not sure but I prefered to cover all the cases.

app/src/main/java/org/akvo/flow/util/SurveyFileNameGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+ * along with Akvo Flow.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package org.akvo.flow.util;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+public class SurveyFileNameGenerator {
```
valllllll2000: I think in this case it is easier to test this one file than a static class which has lots of methods or the BootstrapService. I just wanted to isolate this feature to test it.

app/src/main/java/org/akvo/flow/util/SurveyIdGenerator.java
valllllll2000: I prefer to make sure it also works with more subfolders just in case we never know what may happen, like users adding multiple slashes in the filename, it would then create multiple subfolders.

app/src/main/java/org/akvo/flow/util/SurveyFileNameGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public SurveyFileNameGenerator() {
+    }
+
+    @NonNull
+    public String generateFileName(@Nullable String entryName) {
+        if (entryName == null || entryName.isEmpty()) {
+            return "";
+        }
+        int fileSeparatorPosition = entryName.lastIndexOf("/");
```
valllllll2000: I am not sure this will work because a filename cannot contain '/' see here https://en.wikipedia.org/wiki/Filename

valllllll2000: Ok so I have tried and it does not work. We need to use the filename which is the last bit of '/'. This is because when the bootstrap is generated even if the user created it as a name it is actually made a folder. We should maybe forbid some characters in the actual form name in the backend.

Issue #615, opened 2017-02-02
Fine tuning of sentry settings


valllllll2000:
Sentry has been added and can now be tested using the uat1 build.

There 2 fronts to consider:

1) sentry.io settings 
Privacy has to be taken into consideration: https://docs.sentry.io/learn/sensitive-data/
What information has to be scrubbed?
What can be sent in email notifications and which should not?
2) device settings:
Which information to send? Check if all the tags currently sent are necessary.
Need to be added: deviceId (if available)

muloem:
Just to add a little more: as far as possible lets look into having the scrubbing done at the app side rather than server side.

valllllll2000:
@muloem I have looked at both libraries and I have not found any way to remove sensitive data in either Sentry-Android or raven-android, although raven-android is in current development and they may add it in the future. As for now here is the information that we send:

_data:  example_
appId: akvoflow-uat1
Device: Moto G (4)
device: athene
device.family: motorola
level: error
os: Android 6.0.1
os.name: Android
OsVersion: 6.0.1
Platform: Android
release: 2.2.11
transaction: org.akvo.flow.util.HttpUtil.getStatusCode(HttpUtil.java:136)
VersionCode: 14
VersionName: 2.2.11

valllllll2000:
## Test plan

The test plan is the same as for https://github.com/akvo/akvo-flow-mobile/issues/551

janagombitova:
👍

PR #628, opened 2017-02-17
Sentry tags update (connect #615) 

valllllll2000:
After reviewing all the information sent to sentry.io via tags I have decided to review/update it as some information was repeated
#### The solution
Here are the tags that are going to be sent
**app.id**
akvoflow-uat1
**device** (the library appends it)
angler
**device.family** (the library appends it)
google
**flow.device.id**
valeria_nexus6
**device.model**
Nexus 6P
**level** (the library appends it)
fatal
**os** (the library appends it)
Android 7.1.1
**os.name** (the library appends it)
Android
**os.version**
7.1.1
**release** (the library appends it)
2.2.11
**transaction** (the library appends it)
org.akvo.flow.dao.SurveyDbAdapter.executeSql(SurveyDbAdapter.java:1093)
**version.code**
21
**version.name**
2.2.11
NA
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


muloem:
About the info sent, the **os** and **os.version** are more or less the same, should we drop one?  Same for the **release** and **version.name**

valllllll2000:
@muloem I like having os.name and os.version as separate tags. I also like having version.name and code separate and clearly specified what they are instead of release. I am thinking, then about removing 'release' and 'os' tags from the library itself what do you think?

muloem:
Yes thats probably the way to go then.  Can you remove them by configuration?  Or do you have to make lots of code changes?

valllllll2000:
@muloem Code changes are required in the library repo and then i need to include the latest jar

app/src/main/java/org/akvo/flow/util/logging/TagsFactory.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final String INSTANCE_ID_TAG_KEY = "app.id";
+    private static final String DEVICE_ID_TAG_KEY = "flow.device.id";
+    private static final String DEVICE_MODEL_TAG_KEY = "device.model";
+    private static final String OS_VERSION_TAG_KEY = "os.version";
+    private static final String VERSION_NAME_TAG_KEY = "version.name";
+    private static final String VERSION_CODE_TAG_KEY = "version.code";
+    private static final String DEFAULT_TAG_VALUE = "NotSet";
+    private static final int NUMBER_OF_TAGS = 6;
+
+    private final Map<String, String> tags = new HashMap<String, String>(NUMBER_OF_TAGS);
```
valllllll2000: I think if we add more tags then we change the value. It is recommended for minor performance improvements to always initialise collections with a capacity especially if you know it in advance like in this case.

muloem: So in this case we should initialise it to 10 to avoid having the capacity increased, right?

valllllll2000: there are only 6 tags here, they are added one by one to the existing tags

valllllll2000: ok I have removed the size, thanks for pointing this out

Issue #583, opened 2017-01-09
Improve datapoint sync layout


valllllll2000:
We have decided to improve the usage of notifications as sometimes their use within the app is not correct and does not follow [google design guidelines](https://material.io/guidelines/patterns/notifications.html) at all.
Here is the original issue where all the notifications have been discussed
https://github.com/akvo/akvo-product-design/issues/218

For datapoint sync we have decided on the following specs:

- No notification will be used

- While syncing, there will be a loading spinning wheel (ProgressBar) displayed on top of the datapoint list or map.
- Success or failure will be displayed with a snackbar with some of the error messages will include a retry button. See https://docs.google.com/spreadsheets/d/1ONmVlA9TYHCDEkglfWqIbJse64DNmOyhrl7fFwdEReU/edit for more details and exact text specs.

valllllll2000:
@janagombitova I would need your input on this.
![snackbar](https://cloud.githubusercontent.com/assets/923280/22432615/c5edd578-e715-11e6-8d23-fd7b6f577102.png)
As you can see on the picture, the message is too long for this small device. This happens for both messages:
* Sync failed. Flow dashboard assignment required
* Sync failed. Check your connection and try again

Here are my suggestions:
* Dashboard assignment missing
* Check your connection and retry

janagombitova:
I personally like that the message says something went wrong. So here would be my suggestions: 
- Sync failed. Assignment missing. 
- Sync failed. No connection.

But if we agree that by giving the users the "retry" action, the notion that something went wrong is clear enough then fine. Then I would go for: 
- Dashboard assignment missing
- No connection 

@Kiarii is not available for Flow work so I brought him in the loop on the overall notifications improvement project we started. Thus, Kiarii what are your thoughts? These are the first changes for the orange coloured current notifications.

Kiarii:
If Sync refers to the transfer of data from phone to server, i think it ought to take a simpler approach as you mentioned (with the ticking). 

I also think a Synch failure may occur if the server is down; a situation in which the user cannot intervene from the app. Thus it may be best that the app handles 'Retry' (until success) and not to prevent the user from doing other things or even asking them to initiate the task that the system can/shd handle.

This is my first take, trying to get a good feel on the notifications..

valllllll2000:
@janagombitova I will then be using the strings, they fit in one line perfectly.
Sync failed. Assignment missing.
Sync failed. No connection.

[...]

PR #629, opened 2017-02-20
Datapoint notifications phase1 (connect #583)

valllllll2000:
Notifications are displayed for datapoints sync

#### The solution
We no longer display notifications while syncing datapoints but show a loading ui + snackbar when loading is finished or there were errors.

**The feature is not finished**, this is only the first phase of the feature in order to not have a big PR.
Once all the phases are merged to `datapoint-display-refactor` branch, it can then be merged to develop.

NA
#### Reviewer Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/activity/MapActivity.java
```diff
@@ -19,40 +19,134 @@
...
+        } else {
+            centerMapOnUserLocation();
+        }
+    }
+
+    private void centerMapOnUserLocation() {
+        LatLng position = null;
+        LocationManager manager = (LocationManager) getSystemService(Context.LOCATION_SERVICE);
+        Criteria criteria = new Criteria();
+        criteria.setAccuracy(Criteria.ACCURACY_FINE);
```
valllllll2000: That is a very valid point indeed, i just copied the code that was already there but since it is the user location that we display here it would be totally fine to have coarse location which is faster and uses less battery, I have changed it.

app/src/main/java/org/akvo/flow/activity/MapActivity.java
```diff
@@ -19,40 +19,134 @@
...
+                .findFragmentById(R.id.map);
+        mapFragment.getMapAsync(this);
+        datapointId = getIntent().getStringExtra(ConstantUtil.SURVEYED_LOCALE_ID);
+        loadItem();
+    }
 
-        SurveyDbAdapter db = new SurveyDbAdapter(this).open();
-        SurveyedLocale datapoint = db.getSurveyedLocale(datapointId);
-        db.close();
+    private void loadItem() {
```
valllllll2000: changed to loadDataPoint

app/src/main/java/org/akvo/flow/serialization/response/SurveyedLocaleParser.java
```diff
@@ -73,7 +74,7 @@ public SurveyedLocale parseSurveyedLocale(JSONObject jSurveyedLocale) throws JSO
                 .parseList(jSurveyInstances);
 
         SurveyedLocale surveyedLocale = new SurveyedLocale(id, name, lastModified, surveyGroupId,
-                latitude, longitude);
+                latitude, longitude, SurveyInstanceStatus.SAVED);
```
valllllll2000: I think I will leave this one as is because this class will be removed in phase 4 as the parsing will be completely refactored

valllllll2000:
@muloem I have addressed the issues you commented on

app/src/main/java/org/akvo/flow/ui/fragment/ResponseListFragment.java
```diff
@@ -42,19 +42,18 @@
...
+import static org.akvo.flow.util.ConstantUtil.EXTRA_RECORD_ID;
+import static org.akvo.flow.util.ConstantUtil.EXTRA_SURVEY_GROUP;
+
 public class ResponseListFragment extends ListFragment implements LoaderCallbacks<Cursor> {
     private static final String TAG = ResponseListFragment.class.getSimpleName();
 
-    private static final String EXTRA_SURVEY_GROUP = "survey_group";
-    private static final String EXTRA_RECORD = "record";
-
     // TODO: Move all id constants to ConstantUtil
```
muloem: Is this `TODO` still valid / useful?

valllllll2000: yes it was made by Iñigo and is still relevant but it is used in 2 classes ResponseListFragment and ResponseListAdapter, I can do the changes but I think they are kind of out of scope here since I am mostly working on datapoints and looking at those 2 they will eventually also need refactoring so I would do that at that point, what do you think?

valllllll2000:
@muloem I have updated/replied

Issue #633, opened 2017-02-28
Review some exceptions being sent to sentry


valllllll2000:
Some exceptions are useless and should be removed.
For example lots of useless exceptions come in the line https://github.com/akvo/akvo-flow-mobile/blob/develop/app/src/main/java/org/akvo/flow/util/HttpUtil.java#L61

* java.net.ConnectException
* javax.net.ssl.SSLHandshakeException
* java.security.cert.CertificateNotYetValidException
* javax.net.ssl.SSLProtocolException
* java.net.SocketTimeoutException
* Containing message "Connection timed out"
* https://sentry.io/akvo-foundation/flow-android/issues/223048591/
needs to be removed when forbidden since it is expected

Needs to be improved as for now as it is now it does not help:
https://sentry.io/akvo-foundation/flow-android/issues/222967325/

Make this one more explanotary (add more details):
https://sentry.io/akvo-foundation/flow-android/issues/222966993/

valllllll2000:
## Test plan
I wrote unit tests for the first part (the removed exceptions).
For the rest make sure you have surveys assigned.
Kill and reopen the app and make sure they are synced correctly.
Try syncing a non assigned form (using settings -> download form). 
Try using the app in flight mode (syncing, creating datapoints)
If possible, try to test using 2G networks having the mobile sync settings enabled in preferences.
Try changing the dashboard name using preferences

Write here you device id and device model and I can check in Sentry if these exceptions arrived as expected.

janagombitova:
@valllllll2000 my device ID is: jana and as far as I can tell the app works as it should, can you please cross check if all on Sentry side is as expected as well? I followed your test plan

valllllll2000:
@janagombitova I have checked on sentry and everything is as expected, thanks!

PR #643, opened 2017-03-07
Remove useless exceptions (connect #633)

valllllll2000:
Too many exceptions which were not so useful are sent to sentry
#### The solution
Filter out some of the exceptions
Add test
Improve the information sent by some of the exceptions
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/java/org/akvo/flow/service/DataSyncService.java
+++ b/app/src/main/java/org/akvo/flow/service/DataSyncService.java
@@ -44,6 +44,7 @@ import org.akvo.flow.domain.FileTransmission;
 import org.akvo.flow.domain.Survey;
 import org.akvo.flow.domain.response.FormInstance;
 import org.akvo.flow.domain.response.Response;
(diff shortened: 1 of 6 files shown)


valllllll2000:
@stellanl I have fixed the naming, thanks

app/src/main/java/org/akvo/flow/service/DataSyncService.java
```diff
@@ -591,7 +592,13 @@ private void checkDeviceNotifications() {
                 Timber.e("Could not retrieve missing files");
             }
         } catch (Exception e) {
```
muloem: Small comment here... maybe you could have used instead
```java
try {
} catch (HttpException e) {
} catch (Exception e) {
}
```

Instead of having the `if()` block

Issue #613, opened 2017-01-30
Basic survey question test


MelEnt:
Using expresso, add tests to the following question types:

- [x]  Free text
- [x]  Date
- [x]  Number
- [x]  Option

valllllll2000:
Update: 
FreeTextQuestionTest has already been added
We currently have the branch issue/613-automated-test-1
which contains DateFormTest, NumberSurveyTest, OptionSurveyTest
Tasks left to do:

- [ ] Make sure these tests pass and are deterministic
- [ ] Make sure they work on travis builds

valllllll2000:
Update: most tests work, only left to fix the OptionQuestion test

valllllll2000:
## Test plan
Nothing to test here manually

PR #650, opened 2017-03-13
Automated tests (part 2) (connect #613)

MelEnt:
Tests previously required already installed and set up user. 
Also, this PR adds three new survey types (Date, Number and Options) which is built in the same manner as the previous survey (Freetext)

#### The solution
Proper setup and teardown procedures before and after tests, respectively, will hopefully solve the issue of missing prerequisites.

NA

#### Reviewer Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/androidTest/java/org/akvo/flow/activity/NumberSurveyTest.java
+++ b/app/src/androidTest/java/org/akvo/flow/activity/NumberSurveyTest.java
@@ -55,7 +55,6 @@ import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.endsWith;
 import static org.hamcrest.Matchers.not;
 
-
 public class NumberSurveyTest {
 
     @Rule
@@ -63,10 +62,9 @@ public class NumberSurveyTest {
     private SurveyInstaller installer;
 
     @Before
-    public void init()
-    {
+    public void init() {
         Context context = rule.getActivity();
-        installer       = new SurveyInstaller(context, new SurveyDbAdapter(context));
+        installer = new SurveyInstaller(context, new SurveyDbAdapter(context));
     }
 
     @BeforeClass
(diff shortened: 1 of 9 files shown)


app/src/androidTest/java/org/akvo/flow/activity/NumberSurveyTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+public class NumberSurveyTest {
+
+    @Rule
+    public ActivityTestRule<SurveyActivity> rule = new ActivityTestRule<>(SurveyActivity.class);
+    private SurveyInstaller installer;
+
+    @Before
+    public void init()
+    {
```
valllllll2000: please fix the formatting of {}

app/src/androidTest/java/org/akvo/flow/testhelper/SurveyInstaller.java
```diff
@@ -66,23 +88,25 @@ public Survey persistSurvey(String xml) throws IOException {
...
         adapter.addSurveyGroup(survey.getSurveyGroup());
         adapter.close();
-
         notifyNewSurvey(context);
-        return survey;
+    }
+
+    public void clearSurveys() {
+        for (File file : surveyFiles)
+        {
```
valllllll2000: please fix formatting

valllllll2000:
@MelEnt Please apply the formatting (ctrl + alt + i and ctr + alt + l)  to all your files as the formatting is not correct and some lines are too long.

app/src/androidTest/java/org/akvo/flow/testhelper/SurveyRequisite.java
```diff
@@ -0,0 +1,41 @@
...
+        FlowApp.getApp().setUser(new User(1L, "User"));
+        Prefs prefs = new Prefs(context);
+        //To bypass the need for a setup (create user)
+        prefs.setBoolean(Prefs.KEY_SETUP, true);
+        //To bypass "Low Storage" prompt
+        prefs.setLong(Prefs.KEY_SPACE_AVAILABLE, 50L);
+    }
+
+    //Reset to stage before tests
+    public static void resetRequisites() {
```
valllllll2000: maybe clearRequisites would be a better name

app/src/androidTest/java/org/akvo/flow/activity/OptionsSurveyTest.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    public void canFillOptionsQuestion() throws IOException, InterruptedException {
+        fillOptionsQuestion(optionsurvey, 0);
+        onView(allOf(withClassName(endsWith("Button")), withText(R.string.submitbutton))).check(matches(isEnabled()));
+    }
+
+    private Survey fillOptionsQuestion(int surveyResId, int option) throws IOException {
+        Survey survey = getSurvey(surveyResId);
+
+        openDrawer();
```
valllllll2000: out of curiosity: why do you open the drawer here?

app/src/androidTest/java/org/akvo/flow/testhelper/SurveyRequisite.java
MelEnt: Well, clear would imply they're gone, while reset tells us they're set to the stage before the tests

app/src/androidTest/java/org/akvo/flow/activity/OptionsSurveyTest.java
MelEnt: Before selecting a survey we open the drawer. Even if it's already open, it doesn't hurt to double open it, just for the sake of assurance.

valllllll2000:
@MelEnt @EnderCrypt I have been running all the tests again and the often fail. The reason for that is always the same:
```
android.support.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with id: org.akvo.flow:id/new_datapoint
```
It seems that once you run one of the tests since we are in another activity (FormActivity) the new_datapoint menu button is no longer there so I would separate these tests in 2 parts: 
1) Test the survey activity, selecting survey from drawer and then pressing new_datapoint and make sure we are in FormActivity 
2) The actual tests for diferent question views which should be run using FormActivity started with the required survey.
Currently you seem to be testing the navigation itself multiple times but in my opinion this is not necessary. An alternative would be to maybe press the back button at the end of each test but I would prefer the first suggestion.

Also I have been running your tests with the following emulator config:
`emulator -use-system-libs -avd Pixel_XL_API_25 -no-audio -no-window`
Pixel_XL_API_25 is the name of the emulator
`-use-system-libs` is because my machine has some issues with emulator but you can probably omit that
and also I run the following settings before running the tests:
```
adb shell settings put global window_animation_scale 0 &
adb shell settings put global transition_animation_scale 0 &
adb shell settings put global animator_duration_scale 0 &
adb shell input keyevent 82
```
This is the way to run the tests in travis CI, that is why no window, once you make the fixes we will be able to run your tests using travis which will help us make sure new commits do not break the views.

valllllll2000:
Will update the branch with fixes and create a new PR

PR #638, opened 2017-03-02
Datapoint notifications phase4 (connect #583)

valllllll2000:
#### Before the PR (what is the issue or what needed to be done)
All the logic is inside the fragments and the service which handles the sync sends broadcast receivers to notify the views of new data available.
#### The solution
**This is the last piece of the refactor** 

DataPoint list and map have been refactored to MVP (Model View Presenter) and clean architecture:
* Fragments are in charge of displaying and contain no logic (the exception is the location inside data point list). All user input received will be transmitted to the presenter which decides what to do.
* Presenter contains the logic. The have access to the UseCases which allow to get data in (or data out).
* UseCases are inside the domain module. The contain logic to retrieve data from network, database or shared preferences.

Removed the all datapoint sync service and the broadcast receivers for data point sync.
Using [sqlbrite library](https://github.com/square/sqlbrite) for database access to track changes in the database using [rx-android](https://github.com/ReactiveX/RxAndroid)
#### Screenshots (if appropriate)
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

PR #685, opened 2017-04-03
Datapoint display refactor data networking 2/7 (connect #583)

valllllll2000:
NA
#### The solution
**Phase 2** of the big PR #638 
Added all the base classes for retrieving datapoints from network.
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


data/src/main/java/org/akvo/flow/data/entity/SyncedTimeMapper.java
```diff
@@ -18,11 +18,29 @@
...
-package org.akvo.flow.data.util;
+package org.akvo.flow.data.entity;
 
-public class Constants {
+import android.database.Cursor;
 
-    public static final int IMAGE_SIZE_320_240 = 0;
-    public static final int IMAGE_SIZE_640_480 = 1;
-    public static final int IMAGE_SIZE_1280_960 = 2;
+import org.akvo.flow.database.SyncTimeColumns;
```
muloem: Did you rename this class to something else or is this just the way it was interpreted by git?

valllllll2000: This does not make sense at all, I did not rename Constants to SyncedTimeMapper

data/src/main/java/org/akvo/flow/data/net/DeviceHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        if (teleMgr != null) {
+            number = teleMgr.getDeviceId();
+        }
+        if (number == null) {
+            number = "NO_IMEI";
+        }
+        return number;
+    }
+
+    public String getAndroidID() {
```
valllllll2000: Ok renamed it to getAndroidId(), this method did not exist before

data/src/main/java/org/akvo/flow/data/entity/ApiDataPoint.java
```diff
@@ -0,0 +1,119 @@
...
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        ApiDataPoint that = (ApiDataPoint) o;
+        return id.equals(that.id);
```
valllllll2000: I used the IDE to create these methods but ok, I have changed it to that.getId(), not sure either how come it worked

data/src/main/java/org/akvo/flow/data/net/FlowApiService.java
```diff
@@ -20,6 +20,16 @@
...
+import org.akvo.flow.data.entity.ApiLocaleResult;
 
+import retrofit2.http.GET;
+import retrofit2.http.Headers;
+import retrofit2.http.Url;
+import rx.Observable;
+
+interface FlowApiService {
+
+    @GET
```
muloem: Here its not possible to use the `@Query` annotations to specify what are the query params?

valllllll2000: Yes, this is how it is normally done but our url is a bit odd as some parts are encoded. Will try something tonight and see if it works

valllllll2000: I have tried everything and it just does not work. The way we build the url is: baseUrl + surveyedlocale? + all the query params  + &h=the query prams encoded with HMacSha1, which is not something that can be easily done with retrofit. Also the params have to be ordered alphabetically so if they are reordered it will not work either.

data/src/main/java/org/akvo/flow/data/net/FlowRestApi.java
```diff
@@ -20,12 +20,137 @@
...
+            return "";
+        }
     }
 
+    private String getTimestamp() {
+        SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss", Locale.US);
+        dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
+
+        try {
+            return URLEncoder.encode(dateFormat.format(new Date()), CHARSET_UTF8);
```
muloem: Maybe you can reuse the `encodeParam()` function here?

data/src/main/java/org/akvo/flow/data/net/FlowApiService.java
```diff
@@ -20,6 +20,16 @@
...
 
-public interface FlowApiService {
+import org.akvo.flow.data.entity.ApiLocaleResult;
 
+import retrofit2.http.GET;
+import retrofit2.http.Headers;
+import retrofit2.http.Url;
+import rx.Observable;
+
+interface FlowApiService {
```
muloem: Given that this is focused on syncing data points, would it be more appropriate to call it the `DataSyncService` or something along those lines?  Or do you see this as the definition point for all the API requests going to the backend?

data/src/main/java/org/akvo/flow/data/net/FlowRestApi.java
```diff
@@ -20,12 +20,137 @@
...
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
 import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import rx.Observable;
+import timber.log.Timber;
 
+@Singleton
 public class FlowRestApi {
```
muloem: Similar question here.  Do you see this as the definition class for all API requests or only syncing data points?

data/src/main/java/org/akvo/flow/data/net/FlowApiService.java
muloem: If possible it would be great as it also takes care of the encoding so less code written.

data/src/main/java/org/akvo/flow/data/net/FlowRestApi.java
valllllll2000: This one is for all api calls to the flow api

PR #686, opened 2017-04-03
All the data classes for retrieving datapoints 3/7 (connect #583)

valllllll2000:
NA
#### The solution
Phase 3: datasources for network, database and properties + retrieval of user preferences
does not compile as needs the changes in database module to work
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


data/src/main/java/org/akvo/flow/data/datasource/DatabaseDataSource.java
```diff
@@ -0,0 +1,186 @@
...
+
+    public Observable<List<ApiDataPoint>> syncSurveyedLocales(List<ApiDataPoint> apiDataPoints) {
+        if (apiDataPoints == null) {
+            return Observable.<List<ApiDataPoint>>just(Collections.EMPTY_LIST);
+        }
+        syncSurveyedLocale(apiDataPoints);
+        return Observable.just(apiDataPoints);
+    }
+
+    private void syncSurveyedLocale(List<ApiDataPoint> apiDataPoints) {
```
muloem: I think we can already start moving away from the name `surveyedLocale` and use `datapoint` instead wherever possible

data/src/main/java/org/akvo/flow/data/datasource/DatabaseDataSource.java
```diff
@@ -0,0 +1,186 @@
...
+        if (apiDataPoints == null || apiDataPoints.size() == 0) {
+            return;
+        }
+        BriteDatabase.Transaction transaction = briteSurveyDbAdapter.beginTransaction();
+        try {
+            for (ApiDataPoint surveyedLocale : apiDataPoints) {
```
muloem: same here re `surveyedLocale`

data/src/main/java/org/akvo/flow/data/datasource/DatabaseDataSource.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    /**
+     * JSON array responses are ordered to have the latest updated datapoint last so
+     * we record it to make the next query using it
+     * @param apiDataPoints
+     */
+    private void updateLastUpdatedDateTime(List<ApiDataPoint> apiDataPoints) {
+        ApiDataPoint surveyedLocale = apiDataPoints.get(apiDataPoints.size() - 1);
```
muloem: In case this is an empty list we will get an exception, no?

valllllll2000: I have added a @NonNull annotation and a check for empty list. If you look above it will not be called with empty or null but just in case.

data/src/main/java/org/akvo/flow/data/repository/SurveyDataRepository.java
```diff
@@ -0,0 +1,192 @@
...
+    @Override
+    public Observable<Integer> syncRemoteDataPoints(final long surveyGroupId) {
+        return getServerBaseUrl()
+                .concatMap(new Func1<String, Observable<Integer>>() {
+                    @Override
+                    public Observable<Integer> call(final String serverBaseUrl) {
+                        return dataSourceFactory.getPropertiesDataSource().getApiKey()
+                                .concatMap(new Func1<String, Observable<Integer>>() {
+                                    @Override
+                                    public Observable<Integer> call(String apiKey) {
```
muloem: I don't know yet what is a better way or even if there is one, but the nesting here feels strange for me in terms of code formatting.  If you don't have any idea right away I will merge it but would like to revisit it before the final merge into develop

Issue #748, opened 2017-06-06
Signature question not properly saved in unsubmitted forms


valllllll2000:
**Steps to reproduce**
Create survey with signature question.
Fill the form by also adding the "signed by" text field but do not submit
Press back and see the datapoint has been created
Go back to the datapoint: you will be able to view the signature image but the "signed by" field will be empty.

valllllll2000:
## Test plan:
Create a signature question. When filling the form:
* make sure both name and image are saved
* view saved form
* try clearing saved form, check everything is deleted correctly
* make sure that correctly submitted and can be viewed in read only mode.

janagombitova:
👍  ready to go

Issue #751, opened 2017-06-07
Signature question view design


valllllll2000:
While working on #748 I realized that the name cannot be entered together with the signature which may seem a bit confusing and also the name being underneath the button, it is easy to forget adding it. This little issue contributes to the fact that if the form is saved, only the signature is saved, not the name.  I will attempt to add the possibility to write the name underneath the signature so it will be saved at the same time.

valllllll2000:
Here are the 2 options:
![screenshot_20170607-113138](https://user-images.githubusercontent.com/923280/26872831-3c08b3d8-4b78-11e7-9d1d-ff2f9a1d4fda.png)
![screenshot_20170607-114112](https://user-images.githubusercontent.com/923280/26872841-42ad9064-4b78-11e7-9d80-424ef81759e0.png)

valllllll2000:
and this is how one of them would look like on an old device:
![screenshot_2017-06-07_1153](https://user-images.githubusercontent.com/923280/26872868-57cd1d52-4b78-11e7-9118-5e96b0d0c368.png)

valllllll2000:
and this is how it would look like in the form view:
![screenshot_20170607-120528](https://user-images.githubusercontent.com/923280/26873345-cdafcc4e-4b79-11e7-9fc8-fd7b51de4733.png)

I would make name input not editable in this view. Instead once there is a signature I would change the button text to "edit signature" so user can go there and change the name. What do you think @janagombitova?

janagombitova:
That looks great and can definitely work, not taking away too much space from the signature canvas. I would go with the view where the name input field is in its own line and the descriptive text under it. 

To your second comment: brilliant! Good that you moved the name field above the button. I agree it should only be editable in the separate signature page. And yes changing the button to edit (if a signature) is added it a smart move

Why is the canvas in your last screen shot showing the signature image so small? Previously it was full screen width

How would the validation work? Can I move away from the editor page if I have one of the two (image or name) added?

valllllll2000:
@janagombitova 

> Why is the canvas in your last screen shot showing the signature image so small? Previously it was full screen width
This has always been like this, as this is the view that just displays the signature, not the one where you can draw.

janagombitova:
I thought it was show till the full screen, but I see now on my phone the same. Must have been my mistake.

valllllll2000:
> How would the validation work? Can I move away from the editor page if I have one of the two (image or name) added?

[...]

PR #753, opened 2017-06-07
New layout signature question (connect #751)

valllllll2000:
Signature question layout lacks the name underneath the drawing of the signature
#### The solution
Redesign of the signature question view
also fixes the issue https://github.com/akvo/akvo-flow-mobile/issues/748
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/activity/SignatureActivity.java
```diff
@@ -56,26 +64,59 @@ public void onClick(View v) {
...
+
+    private void updateSendButton() {
+        boolean isNameEmpty = TextUtils.isEmpty(nameEditText.getText().toString());
+        boolean signatureEmpty = mSignatureView.isEmpty();
+        sendButton.setEnabled(!isNameEmpty && !signatureEmpty);
     }
 
     private void clear() {
         mSignatureView.clear();
+        updateSendButton();
```
muloem: Does the `nameEditText` also need to be cleared?

valllllll2000: That's a good point, but this is the button to clear the drawing, I am inclined to leave it like that since you can edit the text some other way, here is a screenshot, what do you think?
screenshot_1497533360

valllllll2000:
@muloem added test plans to both #751 and #748

app/src/main/java/org/akvo/flow/activity/SignatureActivity.java
muloem: I agree. You can clear the text with simply deleting. 👍

Issue #777, opened 2017-06-23
Release information missing in sentry after moving to the new SDK


valllllll2000:
https://sentry.io/akvo-foundation/flow-android/releases/

Since raven-java the release info seems to be mssing

valllllll2000:
## Test plan
This does not really affect the way the app works, not the sending of the crashes, just an extra info which is nice to have in sentry.

janagombitova:
If there is nothing I can test here, then we are ready to go 👍

PR #778, opened 2017-06-23
Sending release info to sentry (connect #777)

valllllll2000:
the release info is missing from latest releases as the previous library was sending it https://sentry.io/akvo-foundation/flow-android/releases/
#### The solution
sending release info (version.name)
NA
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/app/src/main/java/org/akvo/flow/util/logging/RavenEventBuilderHelper.java
+++ b/app/src/main/java/org/akvo/flow/util/logging/RavenEventBuilderHelper.java
@@ -39,10 +39,9 @@ public class RavenEventBuilderHelper implements EventBuilderHelper {
     @Override
     public void helpBuildingEvent(EventBuilder eventBuilder) {
         eventBuilder.withSdkName(RavenEnvironment.SDK_NAME + ":android");
-        String release = tags.get(VERSION_NAME_TAG_KEY);
         for (String key : tags.keySet()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/akvo/flow/util/logging/RavenEventBuilderHelper.java
```diff
@@ -37,6 +41,10 @@ public RavenEventBuilderHelper(Map<String, String> tags) {
     @Override
     public void helpBuildingEvent(EventBuilder eventBuilder) {
         eventBuilder.withSdkName(RavenEnvironment.SDK_NAME + ":android");
+        String release = tags.get(VERSION_NAME_TAG_KEY);
```
muloem: I was just wondering if perhaps it makes sense to include this in the `for` loop below instead
i.e. `if (VERSION_NAME_TAG_KEY.equals()) {...}`

One other issue is whether you want the version name to be included both in the release property and as a tag.

valllllll2000: I have moved it inside the loop. For the second concern: I need both, it is just for practicalities, release info appears in sentry page: https://sentry.io/akvo-foundation/flow-android/releases/ but it is still practical to have the tag inside the issue so you can click it and see all the issues for that tag..

Issue #808, opened 2017-07-13
Navigation drawer refactor


valllllll2000:
In order to work on the https://github.com/akvo/akvo-flow-mobile/issues/670, I need refactor the whole drawer layout first

valllllll2000:
## Test plan
Assign surveys to device.
Make sure surveys are displayed correctly in drawer if you sync surveys automatically and also using bootstrap. Make sure you can select surveys. Make sure all the 3 lower optiins work. Try scrolling navigation drawer (if there are many surveys). Make sure menu gets updated correctly when you select diferent kind of surveys.

janagombitova:
@valllllll2000 when I select Settings, About or Help in the navigation drawer before the selected screen opens I see for a split second the screen with the selected survey and list of data points. It is as if Flow takes a little detour before it brings me to Settings/About/Help

valllllll2000:
@janagombitova 

> And when I am in Settings and I hit on the arrow to go back, I expect to come back to the opened navigation drawer but I come back to the selected survey and list of data poitns


This is the normal behaviour in Android. When you press any entry in the navigation drawer, the drawer first gets dismissed while you get taken to the "other place", pressing back should not take you to navigation drawer. I know it worked diferently in our app untill now but it is not the correct behaviour. I can change it of course if needed.

> when I select Settings, About or Help in the navigation drawer before the selected screen opens I see for a split second the screen with the selected survey and list of data points. It is as if Flow takes a little detour before it brings me to Settings/About/Help

Very similar to my first comment but I do agree the lag seems to be too large, I can reduce it. The reason there is a lag in the first place is sometimes closing the drawer while navigating to another screen sometimes looks "weird" so it is recommended to add a little lag to make it all look smoother.

janagombitova:
@valllllll2000 I learn something new with you every day ;) If all these comments of mine are the proper Android behaviour then lets stick the to patters set and not make our own. They are there for a reason to make the user interaction consistent and thus simple. Keep the lag as it is and also the navigation. 

Ready to roll and thank you for explaining.

Issue #670, opened 2017-03-23
SurveyActivity: new user(s) layout proposal


valllllll2000:
Here are some screenshots of how this could look:
Initial look when user is selected:
![screenshot_1490296091](https://cloud.githubusercontent.com/assets/923280/24265677/a2282ec8-1004-11e7-9dc9-8e4a5042e8d4.png)

after pressing on the spinner, the dropdown is displayed:
![screenshot_1490296096](https://cloud.githubusercontent.com/assets/923280/24265718/bd843d06-1004-11e7-930d-025ffe10ddd1.png)

if you select "+user": (same layout as before)
![screenshot_1490296100](https://cloud.githubusercontent.com/assets/923280/24265736/cb0d9a6c-1004-11e7-8eb8-8511afdb07e2.png)

janagombitova:
@valllllll2000 I have a question looking at how to user part works on your device and comparing it to mine. On my devices when I expand the user part, the entire menu expands and I do not get the little square. I personally prefer that the user part expands within the menu, but this is my opinion. Why is my experience different from yours (different devices and Android versions) and how can we make sure it is unified? 

It looks like this for the newer phone: 
![screen shot 2017-03-22 at 11 21 43](https://cloud.githubusercontent.com/assets/12456965/24193371/54b2f442-0ef2-11e7-8eb9-fda64d094850.png)

For the old dinosaur: 
![screen shot 2017-03-22 at 11 23 13](https://cloud.githubusercontent.com/assets/12456965/24193372/56a77bd8-0ef2-11e7-9491-6abebc88aaff.png)

valllllll2000:
@janagombitova so about the diferences between the new and old devices: there is nothing I can do, the system has diferent dialogs and styling for older devices. 

For the way I am displaying the user, this is a proposal to modify the way the user is displayed, to have to similar to some google apps, for example here gmail:

![screenshot_20170324-180104](https://cloud.githubusercontent.com/assets/923280/24305127/30741806-10bc-11e7-868a-80338c4ee08b.png)
although the behaviour that I display is not exactly the same as from the gmail app so I will try to reproduce it and paste the new screenshots here

valllllll2000:
Here are the new options for displaying the user:
![screenshot_1490377432](https://cloud.githubusercontent.com/assets/923280/24306716/d9ce7b9e-10c1-11e7-837e-f03bde96ff30.png)

After clicking on the user it expands:
[...]

PR #827, opened 2017-08-04
Part 1: navigation drawer refactor (connect #808)

valllllll2000:
Navigation drawer, especially user section needed some layout improvements and logic refactor
#### The solution
Part 1
Added data, database and domain logic for surveys handling 
see #670 for screenshots
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


valllllll2000:
The codacy error cannot be fixed: the empty constructor here is needed for dependency injection

domain/src/main/java/org/akvo/flow/domain/interactor/DeleteSurvey.java
```diff
@@ -0,0 +1,77 @@
...
+    protected <T> Observable buildUseCaseObservable(Map<String, T> parameters) {
+        if (parameters == null || parameters.get(SURVEY_ID_PARAM) == null) {
+            return Observable.error(new IllegalArgumentException("params missing"));
+        }
+        final long surveyToDeleteId = (Long) parameters.get(SURVEY_ID_PARAM);
+        return userRepository.getSelectedSurvey().concatMap(new Function<Long, Observable<?>>() {
+            @Override
+            public Observable<?> apply(Long currentSelectedSurvey) {
+                if (currentSelectedSurvey == surveyToDeleteId) {
+                    return userRepository.clearSelectedSurvey().concatMap(
```
muloem: So what actually happens with the selected survey setting after deletion? Does it pick the next survey in line or just stay empty and does that have some impact as far as what appears in the activity or a crash behind the scenes?

data/src/main/java/org/akvo/flow/data/datasource/preferences/SharedPreferencesDataSource.java
```diff
@@ -68,14 +70,35 @@ public SharedPreferencesDataSource(SharedPreferences prefs) {
         return Observable.just(getString(KEY_DEVICE_IDENTIFIER, DEFAULT_VALUE_DEVICE_IDENTIFIER));
     }
 
-    private int getInt(String key, int defaultValue) {
-        return preferences.getInt(key, defaultValue);
+    public Observable<Long> getSelectedSurvey() {
```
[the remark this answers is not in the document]
valllllll2000: it is the selected surveyId. We keep it to store the current state but also when the app restarts we have to display the selected survey. When user selects another survey, we store it in shared preferences. The current ui does not need it but since we use the current survey to highlight the selected survey in the Navigation drawer, but also in SurveyActivity to display survey name if any and inside the datapoint list to show the right error message (if there is a survey selected or if none is selected)

domain/src/main/java/org/akvo/flow/domain/interactor/DeleteSurvey.java
valllllll2000: When the current survey is deleted we do not highlight anything and display that no survey is selected. This part will be handled in future PRs.

PR #829, opened 2017-08-04
Part 3: navigation drawer refactor (connect #670)

valllllll2000:
This is the third part of the navigation drawer refactor
#### The solution
Added data, database and domain classes for the user part of the navigation drawer
see #670 for screenshots
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


domain/src/main/java/org/akvo/flow/domain/interactor/CreateUser.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.inject.Inject;
+
+import io.reactivex.Observable;
+import io.reactivex.functions.Function;
+
+public class CreateUser extends UseCase {
+
+    public static final String PARAM_USER_NAME = "user";
+    private static final long INVALID_ID = -1L;
```
muloem: Do you see a disadvantage using `SharedPreferencesDataSource.INVALID_ID`?  In the event of a decision to change what this number is, we can keep it consistent..

domain/src/main/java/org/akvo/flow/domain/interactor/CreateUser.java
```diff
@@ -0,0 +1,68 @@
...
+            SurveyRepository surveyRepository, UserRepository userRepository) {
+        super(threadExecutor, postExecutionThread);
+        this.surveyRepository = surveyRepository;
+        this.userRepository = userRepository;
+    }
+
+    @Override
+    protected <T> Observable buildUseCaseObservable(Map<String, T> parameters) {
+        if (parameters == null || parameters.get(PARAM_USER_NAME) == null) {
+            return Observable.error(new IllegalArgumentException("missing user name"));
```
muloem: Translation strings?

domain/src/main/java/org/akvo/flow/domain/interactor/CreateUser.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        String userName = (String) parameters.get(PARAM_USER_NAME);
+        return surveyRepository.createUser(userName)
+                .concatMap(new Function<Long, Observable<Boolean>>() {
+                    @Override
+                    public Observable<Boolean> apply(final Long userId) {
+                        if (userId == INVALID_ID) {
+                            return Observable.error(new Exception("Error inserting user"));
```
muloem: Translation as well here?

domain/src/main/java/org/akvo/flow/domain/interactor/DeleteUser.java
```diff
@@ -0,0 +1,79 @@
...
+    protected <T> Observable buildUseCaseObservable(Map<String, T> parameters) {
+        if (parameters == null || parameters.get(PARAM_USER) == null) {
+            return Observable.error(new IllegalArgumentException("missing user"));
+        }
+        final User user = (User) parameters.get(PARAM_USER);
+        return userRepository.getSelectedUser()
+                .concatMap(new Function<Long, Observable<Boolean>>() {
+                    @Override
+                    public Observable<Boolean> apply(Long selectedUserId) {
+                        if (selectedUserId == user.getId()) {
```
muloem: maybe `selectedUserId.equals()` instead? i.e. they are objects not primitives

domain/src/main/java/org/akvo/flow/domain/interactor/GetUsers.java
```diff
@@ -0,0 +1,78 @@
...
+                .concatMap(new Function<List<User>, Observable<Pair<User, List<User>>>>() {
+                    @Override
+                    public Observable<Pair<User, List<User>>> apply(final List<User> users) {
+                        return userRepository.getSelectedUser()
+                                .map(new Function<Long, Pair<User, List<User>>>() {
+                                    @Override
+                                    public Pair<User, List<User>> apply(final Long selectedUserId) {
+                                        User currentUser = null;
+                                        for (User u : users) {
+                                            if (selectedUserId == u.getId()) {
```
muloem: Ditto `.equals()` comparison.

domain/src/main/java/org/akvo/flow/domain/interactor/CreateUser.java
valllllll2000: CreateUser does not know about SharedPreferenceDataSource and I want to keep them separated. So here I renamed the contsant to INVALID_USER_ID. I do not see any reaosn for the value to change as -1 is the value returned by the database when insert has failed. On the other hand INVALID_ID in SharedPreferenceDataSource can represent any value which is missing and is a long.

domain/src/main/java/org/akvo/flow/domain/interactor/CreateUser.java
valllllll2000: This value is not shown to the user, it is only an exception so no translations here. Domain logic does not need translation, if we were to display something to the user it will be decided in presentation.

domain/src/main/java/org/akvo/flow/domain/interactor/CreateUser.java
valllllll2000: This value is not shown to the user, it is only an exception so no translations here.

Issue #610, opened 2017-01-25
Add proguard when running release builds


valllllll2000:
Proguard makes the apk smaller by removing unused code and libraries.

https://developer.android.com/studio/build/shrink-code.html

Make sure to keep all the needed classes, here is a repo with config files for famous libraries:
https://github.com/krschultz/android-proguard-snippets

Also care must be taken to setup a way in sentry to decode obfuscated stacktrace.

valllllll2000:
Done

PR #853, opened 2017-08-29
Add proguard and remove big library (connect #610)

valllllll2000:
The app cannot be installed on android 2.3 devices due to excess number of methods
#### The solution
Part 1: add proguard + remove big library
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/proguard-files/proguard-rules.pro
```diff
@@ -0,0 +1,56 @@
+# Add project specific ProGuard rules here.
+# By default, the flags in this file are appended to flags specified
+# in /home/valeria/android-sdk-linux/tools/proguard/proguard-android.txt
```
muloem: This is auto generated?  Not sure this has a huge impact on anything but I feel its better to avoid having such a path reference if possible

valllllll2000: yes probably generated automatically, I removed it

Issue #833, opened 2017-08-08
View image in full screen


valllllll2000:
When user clicks on an image in the image question view, the image is supposed to be displayed in a bigger size in a dialog which covers the whole screen.
* This is how it looks on my nexus 6P with android 8.0:
![screenshot_20170808-172137](https://user-images.githubusercontent.com/923280/29080187-6bdf499c-7c5f-11e7-96db-99152ba72c80.png)
![screenshot_20170808-172133](https://user-images.githubusercontent.com/923280/29080190-6f780c42-7c5f-11e7-9a67-dc3f73a4b056.png)
* Here is an example on a sony tablet with android 6.0:
![screenshot_20170808-172353](https://user-images.githubusercontent.com/923280/29080263-a003bd70-7c5f-11e7-9cd8-e1346a604042.png)
* And on a small experia phone with android 2.3.7
![screenshot_2017-08-08_1727](https://user-images.githubusercontent.com/923280/29080570-5eaeb3e2-7c60-11e7-855f-c649940ca680.png)

Question: what is the point in displaying the image just a tiny bit bigger? Should we display the image full screen? 
Bonus: should we maybe allow to zoom in/out on the image?

janagombitova:
We should display the full image if the user clicks on it. I agree the current implementation brings no extra value. Once the image is displayed on full screen do we need to indicate how to close the display or will we rely (as now) on the back navigation of the phone? 

I do not see the point in the bonus. I think full screen display is already a great step forward. But the idea is nice.

valllllll2000:
Here is how it could look
![screenshot_20171009-173029](https://user-images.githubusercontent.com/923280/31345998-df4ff848-ad17-11e7-9a2e-d8a2c73b6359.png)

Not sure what to use as title? options:
"Full screen image"
[Name of the datapoint]
[Name of the form]
?

janagombitova:
Looks good to me. 

We currently show (when capturing new data and also when viewing a submission in History) the Name of the form and the version. I would stick to that for consistency. 

![screenshot_2017-10-11-10-29-45](https://user-images.githubusercontent.com/12456965/31430916-8e94e8a8-ae72-11e7-9fc7-ec3452cf1f07.png)

valllllll2000:
## Test plan
You need survey with image and also one video
Tap the image, make sure it is shown correctly in full screen
Tap video, make sure it is displayed correctly also

janagombitova:
[...]

PR #889, opened 2017-10-12
image full screen (connect #833)

valllllll2000:
When user taps an image it shows a dialog overlay with the image which is a tiny bit bigger but has no use.
#### The solution
Show full screen image
see #833 for screen shots
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/util/ConstantUtil.java
```diff
@@ -133,6 +133,9 @@
     public static final String SURVEY_GROUP_EXTRA = "survey_group";
     public static final String SURVEYED_LOCALE_ID_EXTRA = "surveyed_locale_id";
     public static final String RECORD_ID_EXTRA = "record_id";
+    public static final String IMAGE_URL_EXTRA = "image_url";
+    public static final String FORM_TITLE_EXTRA = "title";
```
muloem: I was a little confused with the name as I though it was referring to forms as we know them, i.e. questions.

valllllll2000: But aren't Surveys actually forms? I used the new name. So this is the name of the form, not the question. I can use the old naming which would make it SURVEY_TITLE_EXTRA

muloem: Ignore the comment. that was my mixup. I just saw now in the issue that you were discussing whether to include the form names or something else. 👍

Issue #620, opened 2017-02-09
Imported data shows "null" for submitter 


janagombitova:
If a user imports new data points via data cleaning, we give the submissions a standard submitter (user name) - IMPORTER. If anyone syncs these data points to the app, in the history of the submission the submitter is shown as `null` this can be confusing. We should maintain consistency and also use `IMPORTER` here. 

![screenshot_2017-02-09-12-19-42](https://cloud.githubusercontent.com/assets/12456965/22781326/4b838f9e-eec2-11e6-908e-72b062927dbd.png)

valllllll2000:
@janagombitova should IMPORTER be translated to other languages or should it be always the same?

janagombitova:
It can always stay the same

valllllll2000:
## Test plan

Create some datapoints with data cleaning tool
Sync datapoints
Open one of the imported ones.
In the history tab you should see IMPORTER as user name.

janagombitova:
looking good 👍

PR #892, opened 2017-10-13
Submitter null when data imported (connect #620)

valllllll2000:
null displayed when no submitter name
#### The solution
use IMPORTER as default name
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/ui/adapter/ResponseListAdapter.java
```diff
@@ -77,11 +79,9 @@ public void bindView(View view, Context context, Cursor cursor) {
 
         String username = cursor.getString(SurveyDbAdapter.FormInstanceQuery.SUBMITTER);
         if (TextUtils.isEmpty(username)) {
-            userView.setVisibility(View.GONE);
-        } else {
-            userView.setVisibility(View.VISIBLE);
-            userView.setText(username);
+            username = DEFAULT_USERNAME;
```
muloem: The view is `VISIBLE` by default then, right?

valllllll2000: yes, the view is visible by default

Issue #752, opened 2017-06-07
Search - no results found


janagombitova:
When the user searches for a data point we show the matching results based on her search. But when se types in something that does not match any of the data points, we do not tell her anything. This can be unclear to some of our users. 

![screen shot 2017-06-07 at 12 27 16](https://user-images.githubusercontent.com/12456965/26875283-72817324-4b81-11e7-8a6f-7a275b9d1c7a.png)

A common patter in to show under the search field "No results found" to tell our user what is happening with her search. 

What do you think? 

Secondly, is it needed to have the Flow icon in the search field? In the image below is how it looks like on the app currently and what the Google material design shows. Its our choice on what to do

![screen shot 2017-06-07 at 12 27 25](https://user-images.githubusercontent.com/12456965/26875282-6ec42a9c-4b81-11e7-9c33-6c3c81faed13.png)

valllllll2000:
@janagombitova 
1) I think we should remove the app icon and replace it with the "back" navigation arrow like in gmail, see screenshot:
![screenshot_20170623-085959](https://user-images.githubusercontent.com/923280/27470096-8488d468-57f2-11e7-9c75-e7e48a75ed13.png)
2) As for when there are no search results I would definitely display the error message: "No results found" in red

valllllll2000:
After discussing with @janagombitova we have decided the following:
+ Search will only be available in the list tab, not in the map
+ Once we confirm the search (enter button in the keyboard), the displayed datapoints list will be filtered and only the datapoints that match the query will show
+ If no datapoints match the query, the message "No results found" will be shown
+ When in searched results screen, pressing back will return us to the "full" non filtered list.

valllllll2000:
Initial view
![screenshot_1498748307](https://user-images.githubusercontent.com/923280/27729081-aef53c2c-5d84-11e7-8e4f-519ea99c82e5.png)

Search button 
![screenshot_1498808591](https://user-images.githubusercontent.com/923280/27729093-b857b07e-5d84-11e7-9035-7b826634e846.png)

janagombitova:
@valllllll2000 I have a question about the view (2nd and 3rd screen shot shared in you last comment) when you start searching for a data point. Do I understand it correctly, that if I type in "2" the list of data points will filter out to show me only those with "2" in their name, but I will also see a list of possible results under the search bar? 

The fact that the search bar has to be in view is ok, if you can make sure the icon matches the style of other icons in Flow. 

[...]

PR #916, opened 2017-10-30
Search fix (connect #752)

valllllll2000:
When making a search, only part of the id is searched
#### The solution
the id has to be returned even if the string searched is not at the beginning of the id
NA
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/domain/src/main/java/org/akvo/flow/domain/interactor/GetSavedDataPoints.java
+++ b/domain/src/main/java/org/akvo/flow/domain/interactor/GetSavedDataPoints.java
@@ -35,6 +35,7 @@ import javax.inject.Inject;
 import io.reactivex.Observable;
 import io.reactivex.ObservableSource;
 import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.Nullable;
 import io.reactivex.functions.Function;
 import io.reactivex.functions.Predicate;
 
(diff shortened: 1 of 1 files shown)


domain/src/main/java/org/akvo/flow/domain/interactor/GetSavedDataPoints.java
```diff
@@ -86,23 +87,27 @@ protected GetSavedDataPoints(ThreadExecutor threadExecutor,
...
                     @Override
                     public boolean test(@NonNull DataPoint dataPoint) throws Exception {
                         String name = dataPoint.getName();
-                        if (name != null) {
-                            name = name.toLowerCase();
-                        }
                         String id = dataPoint.getId();
-                        return ((name != null && name
-                                .contains(filter.toLowerCase()))
-                                || (id != null && id.startsWith(filter)));
```
muloem: Regarding the `id`, I understand that because you dropped this, the search for identifier will also include those where the identifier characters may occur in the middle of the identifier string?  Is this an intentional choice?

valllllll2000: yes, Jana asked for the id to be found even if the string searched is not at the beginning of the id but also in the middle (same behaviour as for name)

Issue #930, opened 2017-11-03
Flow app UI testing


valllllll2000:
The main functionality of the app is the filling out of surveys which is complex and needs tests. The tests that exist are very simple. We need to consider the best approach to full survey ui tests to also to reduce the amount of manual testing needed.
Find a way to run regression tests for the app automatically.
Check https://www.browserstack.com/automate
Check screenshots recording https://github.com/Karumi/KataScreenshotAndroid

valllllll2000:
Question types and constraints

1. Text
2. Option
3. Cascade
4. Number
5. Geolocation
6. Photo
7. Video
8. Date
9. Barcode
10. Geoshape
11. Signature
12. Caddisfly

valllllll2000:
## Test plan
No manual tests required

janagombitova:
👍  this is making me very happy. Ready for release, yay!

PR #951, opened 2017-11-28
Ui tests part5 cascade (connect #930)

valllllll2000:
NA
#### The solution
Added tests for cascade questions
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/ui/view/CascadeQuestionView.java
```diff
@@ -254,7 +254,8 @@ public void resetQuestion(boolean fireEvent) {
         super.resetQuestion(fireEvent);
         updateSpinners(POSITION_NONE);
         if (mDatabase == null) {
-            String error = "Cannot load cascade resource: " + getQuestion().getSrc();
+            String error = getContext()
```
valllllll2000: This is a bit out of scope but I ectracted a hardcoded string so it can also be translated.

Issue #923, opened 2017-11-01
Reload all forms (phase 1)


valllllll2000:
For now in Settings we have the option to reload all forms. It behaves differently than when the forms are reloaded upon app restart. Also the dialogs displayed are too long and there are too many steps. In the issue #660 it has been decided that:
1) Reload all will behave the same way as when the app starts, ie we load all new surveys that have been assigned. We do not delete any surveys.
2) We no longer ask for admin passcode for this operation
3) The dialog:
No title only the question: Reload all forms?
Actions: Cancel, Reload
4) Setting title will be **Reload all forms**

valllllll2000:
@janagombitova the subtitle of that entry is currently: "Deletes and re-downloads assigned surveys (only run if instructed)". Shall we just remove the subtitle as "Reload all forms" says it all.

valllllll2000:
## Test plan
Try downloading 1 form using settings. Make sure form is correctly downloaded.
With the app running and connected to the internet, assign forms to device and use settings to Reload all forms. The behavior should be as described above. All forms should be displayed correctly. Include a cascade in one of the forms and make sure it is downloaded correctly.

janagombitova:
👍  works like a charm

PR #958, opened 2017-12-04
Reverted #923 reload all surveys is back to its original version (connect #923)

valllllll2000:
#### The solution
Reverted the #929 because it prevented partners from recovering from an error state in the app.
NA
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/res/values/strings.xml
```diff
@@ -30,6 +30,8 @@
     <string name="toosmallerr">Value is too small. Minimum value: </string>
     <string name="toolargeerr">Value is too large. Maximum value: </string>
     <string name="baddatatypeerr">Answer must be numeric</string>
+    <string name="reloadsurveyslabel">Reload all surveys</string>
```
muloem: With this revert, there is no need to change other language strings right?

valllllll2000: No, they did not get pulled yet

PR #964, opened 2017-12-16
Forms viewing test (connect #930)

valllllll2000:
#### The solution
Added the empty forms viewing tests class. It tests all the possible form questions to make sure they are displayed correctly. This does not test actual form filling or displaying of filled forms
#### Reviewer Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/app/src/androidTest/java/org/akvo/flow/activity/form/emptyformsview/FormActivityTest.java
+++ b/app/src/androidTest/java/org/akvo/flow/activity/form/emptyformsview/FormActivityTest.java
@@ -1,0 +1,500 @@
+/*
+ * Copyright (C) 2017 Stichting Akvo (Akvo Foundation)
+ *
+ * This file is part of Akvo Flow.
+ *
+ * Akvo Flow is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Akvo Flow is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/akvo/flow/activity/form/emptyformsview/FormActivityTest.java
```diff
@@ -0,0 +1,500 @@
...
+    @AfterClass
+    public static void afterClass() {
+        SurveyRequisite.resetRequisites(InstrumentationRegistry.getTargetContext());
+        installer.clearSurveys();
+    }
+
+    @Test
+    public void viewEmptySurveyTest() {
+        //make sure everything is loaded
+        addExecutionDelay(5000);
```
muloem: I was trying to find out where there is an example to use `IdlingResources` in order to avoid having to sleep.

app/src/androidTest/java/org/akvo/flow/activity/form/emptyformsview/FormActivityTest.java
```diff
@@ -0,0 +1,500 @@
...
+
+        verifyToolBar(survey.getName(), survey.getVersion());
+
+        List<QuestionGroup> questionGroups = survey.getQuestionGroups();
```
muloem: Considering that this is an empty survey shouldn't be verifying something to the effect that there are no groups or questions displayed?

valllllll2000: I just realized the naming is probably misleading. By empty form, I mean a form that has not been filled in, ie, adding the datapoint not viewing it. I will rename it

app/src/androidTest/java/org/akvo/flow/activity/form/emptyformsview/FormActivityTest.java
valllllll2000: From what I understand you need to implement them in app code which I do  not like. I will try removing the execution delay and see if it works.

Issue #959, opened 2017-12-04
Form files seem to have been deleted from the device


valllllll2000:
This issue has been happening across multiple release versions to many devices and many dashboards and it has been reported by sentry https://sentry.io/akvo-foundation/flow-android/issues/300021634/.

A temporary fix is to reload all surveys using the settings -> reload all surveys (see #953 )

We need to find the reason why the files have been deleted. Possible reasons:
* File was deleted by user for some reason
* File was deleted by the system
* No space on device
* Delete all data is pressed (but this also deletes users, forms, images, data)
* Form download failed
* Forms can be deleted after sync if the form has been unassigned from a given device

We need to find out if we are deleting the forms by mistake in some cases

valllllll2000:
https://sentry.io/akvo-foundation/flow-android/issues/300021634/

valllllll2000:
https://sentry.io/akvo-foundation/flow-android/issues/304693248/

valllllll2000:
https://sentry.io/akvo-foundation/flow-android/issues/292051062/

valllllll2000:
## Ideas on how to reproduce the issue
Connect device to computer via usb.
On the device, if possible select something like "Mass Storage" in the usb options (not all devices allow to change.
Open the folder ....`/Android/data/org.akvo.flow/` with the computer.
The issue happens more often on samsung devices. It happens more often with android 6.0.1.
If the device has an external SD card slot, inserting and sd card may help.
Restarting the device may help.
Disconnecting the device without selecting "safely remove" may help.

The issue happens when the device file system is disrupted in some way. So the way to try is:
1) Try getting all the surveys with usb connected, try adding datapoint, it should work
2) Disconnect the usb cable, try adding datapoint again, this time it may fail
If the forms get "corrupted" do not try to recover, wait for the fix to make sure the fix fixes the issue.

valllllll2000:
The fix has been to move the surveys to use app internal storage which should always be available whereas external may not in some cases. It will not fix an issue already existing on devices but will prevent future occurrences of it on newly installed surveys. The code also looks in multiple places for existing surveys.

## Test plan:
Create a survey with at least a mandatory cascade question. Please tests on new and old devices.
* Part 1
Download previous app version
Make sure your survey is downloaded correctly and you are able to create datapoints.
Update the app.
Make sure you can still create datapoints

* Part 2
Clean download of latest version.
Put phone in flight mode
[...]

PR #966, opened 2017-12-18
Move surveys to internal storage (connect #959)

valllllll2000:
We used external app storage to storage forms and their resources. The issue with that is the external storage may sometimes be unavailable. see https://developer.android.com/training/data-storage/files.html#WriteExternalStorage
#### The solution
Move forms to internal storage but still make sure we look for forms in the old location for forms installed with previous app versions.
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/api/S3Api.java
```diff
@@ -62,7 +62,7 @@
     private String mAccessKey;
     private String mSecret;
 
-    public S3Api(Context c) {
+    public S3Api() {
```
valllllll2000: Codacy was complaining about it

Issue #981, opened 2018-01-11
Confirm overwriting geolocation data 


janagombitova:
Created via Reamaze: https://flowhelp.reamaze.com/admin/conversations/enhancement-geolocation-question-in-the-app

# Context 
A common practice among data collectors is verification of data before submitting. This means that after the interaction with the respondent is complete, they go through the questionnaire once again to ensure every thing is ok before submitting , since the data is not editable post submitting.

Generally this task is done in bulk, say during the afternoon or end of day.

During this verification, a common mistake happens with the GPS questions. People inadvertently click the check geolocation button while scrolling, many times by accident. This leads to their current GPS location being recorded and not the respondents.

For all other types of questions mistakes because of inadvertent touches can always be undone, but not the geolocation question.

# Opportunity 
If there is already a location recorded for a geolocation question, then on subsequent clicks of the “Check Geolocation” button, show a warning dialog before overwriting.

valllllll2000:
@janagombitova which text should we display? How about:
Title: Reset and update collected gps coordinates?
Buttons:
*Reset
*cancel

janagombitova:
How about: 
Update GPS coordinates? 
Update
Cancel

or 

Change GPS coordinates? 
Change
Cancel

valllllll2000:
Update sounds better to me:
Update GPS coordinates?
Update
Cancel

valllllll2000:
## Test plan
Create a geolocation question preferably one you can edit manually. Fill in data or acquire location automatically. Once you have a latitude and lognitude press "check geolocation" again. You should see a dialog requesting confirmation.
![screenshot_1516127271](https://user-images.githubusercontent.com/923280/35005352-6723ad72-faf3-11e7-99d3-c8fc91cfe983.png)

muloem:
Just as a comment for a longer term solution, perhaps we want to add the _Review_ state to a form? i.e. after collecting and saving, its viewable and only if we select edit, are the responses editable.  That way inadvertently clicking edit, not only for the GPS but for all other questions is avoided.

valllllll2000:
@muloem I agree that it would be a better solution but it will take a significant amount of work to add a new state so how about we just add this change for now and I create an issue with your idea and we can think about it further? As you can see mentioned in the issue above mistakes for other question types can be easily undone. What do you both think @janagombitova @muloem?

janagombitova:
[...]

PR #983, opened 2018-01-16
Confirm geo overwrite (connect #981) 

valllllll2000:
GPS data may be reset by error if the user presses the button by mistake.

#### The solution
If there are already existing gps coordinates, display a dialog to confirm user wants to delete existing coordinates and get new ones.

screenshot_1516127271

#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/java/org/akvo/flow/activity/FormActivity.java
+++ b/app/src/main/java/org/akvo/flow/activity/FormActivity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2010-2017 Stichting Akvo (Akvo Foundation)
+ * Copyright (C) 2010-2017,2018 Stichting Akvo (Akvo Foundation)
  *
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/akvo/flow/activity/FormActivity.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2010-2017 Stichting Akvo (Akvo Foundation)
+ * Copyright (C) 2010-2017,2018 Stichting Akvo (Akvo Foundation)
```
muloem: this is technically correct. Note that since it was last edited in 2017 the shorthand version 2010-2018 expresses the same thing

Issue #1038, opened 2018-03-15
Settings to temporary publish data - Logic


valllllll2000:
Data and domain logic to handle pusblishing of data and media files

valllllll2000:
## Test plan
Nothing to test here

janagombitova:
👍

PR #1039, opened 2018-03-15
Part 6: Publish data logic (connect #1038)

valllllll2000:
Part 6 of securing collected data
#### The solution
Added alarm manager to track when we need to unpublish files
Added Receiver to track device reboot.
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


janagombitova:
@muloem the user will only interact with the word 'Publish' https://user-images.githubusercontent.com/923280/37451250-50867a02-2831-11e8-9905-d424ae0b4afa.png and https://user-images.githubusercontent.com/923280/37451249-504f8a60-2831-11e8-990f-a8e67ce69d80.png 

Flow takes care of un-publishing automagically for the user, so she does not have to worry about having data available in the publish file once she uploaded it. 

We decided to use the same word as with publishing forms, as in Flow world we mean "make this ready to do something"- make the form ready for data collection, make the data captured offline ready for bulk upload. 

The users we tested with understood the word and its function right away. 

So in my opinion, it should be ok.

Issue #1056, opened 2018-03-27
Improve user handing in FlowApp


valllllll2000:
Keeping the user as var in FlowApp and accessing it from everywhere is not the correct way to do it. In order to work on #1049 I first need to fix this dependency. The work includes:
Creating UseCase to get current user.
Make sure FlowApp does not keep reference to user.
Add a way to all the activities to access the user via the UseCase

valllllll2000:
## Test plan
This should be covered while testing "delete all" in settings but make sure to test user operations just incase:
add user, select it, delete user, make sure cannot delete current user. Put app in background, make sure the same user is still seelcted

janagombitova:
User management works like a charm 👍 Ready to roll

PR #1057, opened 2018-03-27
Improve current user access (connect #1056)

valllllll2000:
User was stored in FlowApp and it had to be accessed through many places which prevented me from making the changes required in settings/delete everything
#### The solution
Access current user with UseCase
Remove user field from FlowApp
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/activity/RecordActivity.java
```diff
@@ -114,22 +121,39 @@ public void onSurveyClick(final String formId) {
...
+        }, null);
+
+
+    }
+
+    private void displayForm(User user, String formId, Survey survey) {
         // Check if there are saved (non-submitted) responses for this Survey, and take the 1st one
-        long[] instances = mDatabase.getFormInstances(recordId, formId,
-                SurveyInstanceStatus.SAVED);
+        long[] instances = mDatabase.getFormInstances(recordId, formId, SurveyInstanceStatus.SAVED);
```
muloem: I guess these are automatic changes by the IDE.  remember `git add -p` to exclude unless needed.

app/src/main/java/org/akvo/flow/activity/SurveyActivity.java
```diff
@@ -455,12 +475,26 @@ private void displayForm(String surveyedLocaleId, User user) {
                 formInstanceId, readOnly, mSurveyGroup);
     }
 
-    private void displaySelectedUser() {
-        User user = FlowApp.getApp().getUser();
-        if (user != null) {
-            Toast.makeText(this, getString(R.string.logged_in_as) + " " + user.getName(),
-                    Toast.LENGTH_LONG).show();
-        }
+    private void verifySelectedUser() {
```
muloem: I was trying to understand why you changed this method to `verifySelectedUser()` because it seems to me the main objective stays the same., i.e. display the user, but before displaying we verify that the username exists.

app/src/main/java/org/akvo/flow/activity/RecordActivity.java
valllllll2000: I think it makes sense to include this change in here as creating a specific PR for just formatting seems an overhead for me and since I have changed this file it makes sense to include it here as well.

app/src/main/java/org/akvo/flow/activity/SurveyActivity.java
valllllll2000: Ok, you are right I will rename it back

Issue #1049, opened 2018-03-21
Delete collected data and published files


valllllll2000:
When user pressed delete collected data or delete everything, we should also delete published files and reset all the settings related to published files.

valllllll2000:
## Test plan
For all these actions you should be required to provide a passcode
In settings, try deleting collected data -> collected data and published files should be deleted
In settings, try deleting all data -> collected data and published files should be deleted and you should be taken to login screen. After logging in you will see that the surveys are downloaded again

PR #1066, opened 2018-03-30
Delete collected data setting UI part 1 (connect #1049)

valllllll2000:
UI needs to updated to use the new logic for deleting data.
#### The solution
Part 1: added all the dialogs related to passcode request
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/presentation/settings/passcode/PassCodeDialog.java
```diff
@@ -0,0 +1,90 @@
...
+        builder.setMessage(R.string.authtext);
+        final EditText input = (EditText) main.findViewById(R.id.pass_code_et);
+        builder.setView(main);
+        builder.setPositiveButton(R.string.okbutton, new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                String val = input.getText().toString();
+                if (ConstantUtil.ADMIN_AUTH_CODE.equals(val)) {
+                    onPassCodeCorrect();
+                    if (dialog != null) {
```
muloem: We are dismissing the dialog regardless of which if branch so maybe it can be a standalone if branch?

app/src/main/java/org/akvo/flow/presentation/settings/passcode/PassCodeDialog.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+        });
+
+        builder.setNegativeButton(R.string.cancelbutton,
+                new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int which) {
+                        dismiss();
+                    }
+                });
```
muloem: formatting seems a bit off here.

app/src/main/java/org/akvo/flow/presentation/settings/passcode/PassCodeDialog.java
valllllll2000: You are so right, fixed!

app/src/main/java/org/akvo/flow/presentation/settings/passcode/PassCodeDialog.java
valllllll2000: Actually it is correct, I have tried formatting shortcuts and it correct

PR #1068, opened 2018-04-02
Delete collected data setting UI part 3 (connect #1049)

valllllll2000:
NA
#### The solution
Part 3 for deleting collected data, using the logic in the UI.
Removed old classes no longer used
Using the previously added dialogs
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.java
```diff
@@ -136,5 +149,72 @@ public void destroy() {
...
+    public void deleteCollectedData() {
+        unSyncedTransmissionsExist.execute(new DefaultObserver<Boolean>(){
+            @Override
+            public void onError(Throwable e) {
+                Timber.e(e);
+                view.showDeleteCollectedData();
+            }
+
+            @Override
+            public void onNext(Boolean exist) {
```
muloem: Why not a primitive here? Also maybe use the convention of prefixing with `is...` or `has...`?

app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.java
```diff
@@ -136,5 +149,72 @@ public void destroy() {
...
+    public void deleteAllData() {
+        unSyncedTransmissionsExist.execute(new DefaultObserver<Boolean>(){
+            @Override
+            public void onError(Throwable e) {
+                Timber.e(e);
+                view.showDeleteAllData();
+            }
+
+            @Override
+            public void onNext(Boolean exist) {
```
muloem: Same as previous comment of Boolean

app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.java
```diff
@@ -136,5 +149,72 @@ public void destroy() {
...
+    }
+
+    private class ClearDataObserver extends DefaultObserver<Boolean> {
+        @Override
+        public void onError(Throwable e) {
+            view.showClearDataError();
+        }
+
+        @Override
+        public void onNext(Boolean cleared) {
```
muloem: Ditto

app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.java
valllllll2000: 1) We are looking if unsynced transmissions exists or not, I cannot come up with a better name except, isExisting?

2) rxjava uses Boolean better use it. This how it is in all Observers

app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.java
valllllll2000: rxjava uses Boolean better use it. This how it is in all Observers

app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.java
valllllll2000: rxjava uses Boolean better use it. This how it is in all Observers

Issue #1106, opened 2018-05-14
Publish data not available when minute 0 is still running 


janagombitova:
When testing 2.5.0 I noticed that after the 90 min mark the Publish data did not refresh to be clickable again if the remaining time was 0 minutes. I assume the timer was still counting down the remaining seconds, but to me it was confusing on why I cannot publish again. 

Eventually the publish updated. 

How might we make the last 0 minute more understandable to the user? 

![screenshot_2018-05-14-10-15-23](https://user-images.githubusercontent.com/12456965/39985963-8a8f6bce-5760-11e8-9ebd-1d6c6ef65088.png)

janagombitova:
I like option 1. It does not complicate the implementation, nor leave us with 0 minutes.

janagombitova:
(Y) works like a charm

PR #1109, opened 2018-05-14
[#1106] Displaying remaining time left approximately rounding up

valllllll2000:
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/presentation/settings/publish/PublishFilesPreferencePresenter.java
```diff
@@ -100,7 +100,7 @@ public void onNext(Long publishTime) {
                         .calculateTimeSincePublished(publishTime);
                 if (timeSincePublished < PublishedTimeHelper.MAX_PUBLISH_TIME_IN_MS) {
                     int remainingTime = publishedTimeHelper.getRemainingPublishedTime(
-                            timeSincePublished);
+                            timeSincePublished) + 1;
```
muloem: Since we have the `PublishedTimeHelper` shouldn't the +1 be in that class instead?  Otherwise if it happens that you want remaining published time in a different class you would also need to add the +1 in that other class.  And maybe add the +1 either with a comment or as a constant with a descriptive name. 👍

valllllll2000: OK, you are right, I fixed it.

Issue #1084, opened 2018-04-18
Transmission keeps absolute path to files


valllllll2000:
The Transmission table saves the full path to the files on the device to the database but this is not recommended. We should use only the filename and folder and rebuild the path as it may change.

valllllll2000:
## Test plan
with version 2.4.8  add some datapoints and make sure they are synced
update to latest version
check datapoint transmission - > path should be filename only

kymni:
Looks good

valllllll2000:
tested on both android 6.0 and 8.0 devices and works as expected

PR #1134, opened 2018-06-14
Fix transmission paths (connect #1084)

valllllll2000:
Transmission contains absolute full path to media and zip files but this can lead to issues if the file has been moved, in applications with several users... 
#### The solution
We need to save only the file name and rebuild the path since we know where the files are placed. We also need to:
Migrate old transmissions to make sure they contain only filename
Make sure datasync logic uses the new path
Fix tests related to the database migration
When files are missing from the dashboard, we get their full paths, make sure to search only the filename in our database.
Stop updating transmissions after moving private files to public folder
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,8 +48,9 @@ sentry.properties
 
 !/util/upload-apk/build/libs/deploy-1.0.jar
 
-#fastlane
+#CI config
 fastlane/report.xml
 fastlane/README.md
 build_config
 .env.default
+app/fabric.properties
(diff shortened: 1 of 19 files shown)


.gitignore
```diff
@@ -48,8 +48,9 @@ sentry.properties
 
 !/util/upload-apk/build/libs/deploy-1.0.jar
 
-#fastlane
+#CI config
 fastlane/report.xml
 fastlane/README.md
 build_config
 .env.default
+app/fabric.properties
```
valllllll2000: just had to add it as it was very annoying

valllllll2000:
The errors shown in codacy do not make sense

Issue #1234, opened 2018-10-11
Read Phone State Permission


valllllll2000:
This permission is quite sensitive and users do not like it.
If we request for it we need to explain why.
Here is where it is used:
Phone number -> only below android 21 
Imei -> all devices
This permission is need right at the beginning of opening the app as we use the phone info to forms to the device

muloem:
Is it possible to run things without this permission if it is refused?

valllllll2000:
@muloem there is no simple answer:
* Devices with android 5.0 -> they cannot refuse

* Devices with android 6.0:
1) For new installs we could remove the usage of imei and phone, we still have to find a a way to know if it's a new install
2) For existing ones: we need to check because imei is sent when getting assignments, datapoints, syncing data. Not sending it, would require the users to be logged out and all assignments to be reassigned unless we do some changes in the backend.

valllllll2000:
## Test plan
With android 6.0 device
Have surveys assigned 
Update the app from previous version using beta
Make sure you are asked for permissions and accept them
Delete everything form phone
Login again and make sure the correct surveys are downloaded
Make sure you are able to submit data correctly

PR #1239, opened 2018-10-24
[#1234] Using DeviceHelper + checking for permissions to avoid crashes (Connect #1234)

valllllll2000:
phone permissions not handled
#### The solution
part 1: check for permission to avoid crashes, this is the first part of the fix
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -2,7 +2,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="org.akvo.flow">
 
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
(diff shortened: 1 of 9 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="org.akvo.flow">
 
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
```
valllllll2000: Just removed useless permission since we have device Write it also grants read

Issue #1252, opened 2018-11-07
Media Uri permissions


valllllll2000:
With api 24 we need to start using FileProvider for shared uri.

valllllll2000:
## Test plan
TEst filling image, video and signature questions

PR #1256, opened 2018-11-07
Issue/1252 uri permission (connect #1252)

valllllll2000:
uris not handled properly for new apis
#### The solution
using FileProvider. after moving to file provider turns out image were duplicated in the DCIM folder again on some devices so had to add logic to remove them. Also logic for images handling needed a bit of improvement: extracted some helper classes.
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


data/src/main/java/org/akvo/flow/data/datasource/files/BitmapHelper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2018 Stichting Akvo (Akvo Foundation)
+ * Copyright (C) 2018 Stichting Akvo (Akvo Foundation)
```
valllllll2000: So this is just very odd, I moved some code from ImageDataSource to BitmapHelper and also moved them to another package and somehow github got it all mixed up.

Issue #1297, opened 2019-01-28
Add tests to data syncing logic


valllllll2000:

valllllll2000:
## Test plan
Please make sure data syncing works as expected while submitting, pressing upload manually, also from app settings.

PR #1302, opened 2019-02-04
Issue/1297 part3 (connect #1297)

valllllll2000:
Data syncing logic needs to be tested
#### The solution
Part 3: added tests to mappers and getSurveyTransmissions method
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/data/src/test/java/org/akvo/flow/data/entity/S3FileMapperTest.java
+++ b/data/src/test/java/org/akvo/flow/data/entity/S3FileMapperTest.java
@@ -1,3 +1,23 @@
+/*
+ * Copyright (C) 2019 Stichting Akvo (Akvo Foundation)
+ *
+ * This file is part of Akvo Flow.
+ *
+ * Akvo Flow is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Akvo Flow is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


data/src/test/java/org/akvo/flow/data/entity/S3FileMapperTest.java
```diff
@@ -0,0 +1,168 @@
...
+        S3File file = s3FileMapper.transform(null);
+
+        assertNull(file);
+    }
+
+    @Test
+    public void shouldReturnNullForEmptyFilename() {
+        S3FileMapper s3FileMapper = new S3FileMapper(mockFileBrowser, mockFileHelper);
+
+        S3File file = s3FileMapper.transform(null);
```
muloem: Since the test is for an empty filename I was expecting this to be `transform("")`?

data/src/test/java/org/akvo/flow/data/entity/TransmissionMapperTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void shouldReturnEmptyForNullCursor() {
+        TransmissionMapper mapper = new TransmissionMapper(null);
+
+        List<Transmission> transmissions = mapper.transform(null);
+
+        assertTrue(transmissions.isEmpty());
+    }
+
+    @Test
+    public void shouldReturnEmptyForEmptyCursor() {
```
muloem: Same here.. is there no difference between null and empty?

PR #1329, opened 2019-02-21
Issue/1328 form download tests

valllllll2000:
form downloading logic refactored
#### The solution
Add tests to the new logic for downloading forms
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


data/src/test/java/org/akvo/flow/data/entity/FormHeaderParserTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+        assertNotNull(result);
+        assertEquals("33", result.getRegistrationSurveyId());
+    }
+
+    @Test
+    public void parseOneShouldReturnCorrectValuesIfStartsWithoutComa() {
+        FormHeaderParser formHeaderParser = new FormHeaderParser();
+
+        ApiFormHeader result = formHeaderParser.parseOne("1,cde,abc,cde,6.0,cde,true,33");
```
muloem: This one I dont get very well.  So even if there is no comma at the beginning the parsing should still identify correctly the format? is it because the format is different for some forms? Are there cases with the comma where something comes before the comma that is not the `id`?

data/src/test/java/org/akvo/flow/data/repository/FormDataRepositoryTest.java
```diff
@@ -0,0 +1,245 @@
...
+    Cursor mockCursor;
+
+    private MockWebServer mockWebServer;
+    private FormDataRepository formDataRepository;
+    private RestApi restApi;
+
+    @Before
+    public void setUp() throws IOException {
+        mockWebServer = new MockWebServer();
+        mockWebServer.start(8080);
```
muloem: Do you actually start up a server or is it just setting the port and all the responses are mocked?

data/src/test/java/org/akvo/flow/data/entity/FormHeaderParserTest.java
valllllll2000: yes the first item is either empty or it can be the phone number, not sure why it's like this.

data/src/test/java/org/akvo/flow/data/repository/FormDataRepositoryTest.java
valllllll2000: it starts a server https://github.com/square/okhttp/blob/master/mockwebserver/src/main/java/okhttp3/mockwebserver/MockWebServer.java#L344

data/src/test/java/org/akvo/flow/data/entity/FormHeaderParserTest.java
valllllll2000: https://github.com/akvo/akvo-flow/blob/ffac2098b0e5b66471b65c28aaa3f62999531682/GAE/src/org/waterforpeople/mapping/app/web/SurveyManagerServlet.java#L141

valllllll2000: https://github.com/akvo/akvo-flow/blob/ffac2098b0e5b66471b65c28aaa3f62999531682/GAE/src/org/waterforpeople/mapping/app/web/SurveyManagerServlet.java#L216

Issue #1335, opened 2019-03-06
Remove old caddisfly implementation


valllllll2000:
Currently on uat1, you can still enable the "Allow external sources" which was the old caddisfly implementation. In the app this enables a button but then the caddisfly app is not able to process these requests any more so there is a message saying no app can handle the request. With Jana we have decided to remove this feature since it is no longer used. 
dashboard issue: https://github.com/akvo/akvo-flow/issues/2998

valllllll2000:
## Test plan
Make sure text question work as expected

PR #1336, opened 2019-03-06
[#1335] Removed old caddisfly implementation (connect #1335)

valllllll2000:
old caddisfly implementation still displayed but the question cannot be answered
#### The solution
remove the old caddisfly implementation, text question can only be text
NA
#### Reviewer Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -97,7 +97,6 @@
     <string name="error_answer_match">Answers do not match</string>
     <string name="error_img_preview">Can\'t load image preview</string>
     <string name="error_video_preview">Can\'t load video preview</string>
-    <string name="use_external_source">Use External Source</string>
     <string name="caddisfly_test">Go to test</string>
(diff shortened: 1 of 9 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -97,7 +97,6 @@
     <string name="error_answer_match">Answers do not match</string>
     <string name="error_img_preview">Can\'t load image preview</string>
     <string name="error_video_preview">Can\'t load video preview</string>
-    <string name="use_external_source">Use External Source</string>
```
muloem: Do you also need to remove these for other languages?

muloem:
Oh and there are conflicts to be sorted out before merging

app/src/main/res/values/strings.xml
valllllll2000: The other languages will be removed when I pull the latest strings from transifex. The idea is to only modify the main strings.xml and let transifex API modify the others, transifex overrides the translation files anyway.

PR #1373, opened 2019-04-16
Issue/1360 settings tracking

valllllll2000:
NA
#### The solution
Adding tracking to settings screen to find out how collectors use the app and which settings are useful
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/presentation/settings/PreferenceActivity.java
```diff
@@ -207,44 +222,67 @@ void onDownloadFormOptionTap() {
...
+                trackingHelper.logLanguageChanged(languages.get(position));
+            }
             presenter.saveAppLanguage(position, languages);
         }
     }
 
     @OnItemSelected(R.id.preference_image_size)
     void onImageSizeSelected(int position) {
-        presenter.saveImageSize(position);
+        if (trackChanges) {
```
muloem: Only here I see that the checking for whether tracking is enabled before logging an event. It isn't clear to me how that checking is handled in other places.

app/src/main/java/org/akvo/flow/presentation/settings/PreferenceActivity.java
```diff
@@ -207,44 +222,67 @@ void onDownloadFormOptionTap() {
...
+            if (trackingHelper != null) {
+                trackingHelper.logImageSizeChanged(position);
+            }
+            presenter.saveImageSize(position);
```
muloem: Shouldn't this be outside of the `if (trackChanges) {}` statement?  i.e. it should happen whether tracking is on or not no?

app/src/main/java/org/akvo/flow/presentation/settings/PreferenceActivity.java
valllllll2000: Oh this had nothing to do with tracking, I have renamed it

app/src/main/java/org/akvo/flow/presentation/settings/PreferenceActivity.java
valllllll2000: Oh this had nothing to do with tracking, I have renamed it

Issue #1342, opened 2019-03-13
Offline area creation


valllllll2000:
Allow selection of an offline map

<img src="https://user-images.githubusercontent.com/923280/54287420-e840ec00-45a5-11e9-91c7-695d4896cbdc.png" width="200">

PR #1397, opened 2019-05-20
[#1342] Offline maps download part 1

valllllll2000:
NA
#### The solution
In this issue I add the possibility to select offline areas on a map to download them. This is the first part which adds all the resources, the changes in the manifest (unfortunately there was also some formatting change) and a mapper.
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/AndroidManifest.xml
```diff
@@ -1,177 +1,187 @@
...
+        </activity>
+        <activity
+            android:name=".presentation.datapoints.map.offline.selection.OfflineMapDownloadActivity"
+            android:label="@string/offline_map_activity_title" />
+
         <meta-data
-                android:name="io.fabric.ApiKey"
-                android:value="89dc21727e76a10511ecda549cab1b2d3aff2ec7"/>
+            android:name="io.fabric.ApiKey"
+            android:value="89dc21727e76a10511ecda549cab1b2d3aff2ec7" />
```
muloem: its not problematic to have this key in here I guess no? i.e. security issue

valllllll2000: That's how they tell you to do it in the documentation, we could extract it to string resources but in the end it does not change much. This key is used to send betas to fabric, not sure what else anyone could do with it.

Issue #1406, opened 2019-06-14
External result is lost if activity is destroyed due to low memory


nhternup:
Original issue: https://github.com/akvo/akvo-caddisfly/issues/235

On devices with low memory, it seems that Android sometimes destroys the Form activity in Flow app. When this happens, the result returned to the Flow app from Caddisfly or other external source is lost as the restored Form activity may not remember the question to associate the result to.

### Steps to reproduce

1. Go to developer settings on the Android device
2. Turn on 'Don't keep activities'
3. In Flow app, launch a Caddisfly test question (Example: Del Agua - Turbidity Tubes)
(Note: Try with a Caddisfly test that does not require a photo to be taken as that also currently has the same issue and will be fixed in the next version.)
5. Input a result in Caddisfly and submit result
6. Note that the result does not appear in Akvo Flow

### Possible solution

Save the question details in onSaveInstanceState and restore from savedInstanceState

PR #1407, opened 2019-06-17
[#1406] Save & restore question details, Set result after tab is loaded

nhternup:
**Fixes** #1406 & https://github.com/akvo/akvo-caddisfly/issues/235

On low memory devices, Android sometimes kills the calling activity. Due to this, the result returned from the external activity could be lost. The question details needed to be saved and restored so that the result could be associated with the question correctly.

#### The solution
1. Save the question details in onSaveInstanceState and restore from savedInstanceState
1. Also wait for Survey Tab to be reloaded before setting the result

NA

#### Test Plan
1. Turn on 'Don't keep activities' in Android Developer settings
1. Test that Caddisfly, Photo, Video, Gallery, Barcode, Shape & Signature questions work correctly

#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/java/org/akvo/flow/activity/FormActivity.java
+++ b/app/src/main/java/org/akvo/flow/activity/FormActivity.java
@@ -177,6 +177,11 @@ public class FormActivity extends BackActivity implements SurveyListener,
         languageMapper = new LanguageMapper(context);
         surveyLanguagesDataSource = new SurveyLanguagesDbDataSource(context);
 
+        if (savedInstanceState != null) {
+            mRequestQuestionId = savedInstanceState.getString(ConstantUtil.REQUEST_QUESTION_ID_EXTRA);
+            imagePath = savedInstanceState.getParcelable(ConstantUtil.IMAGE_FILE_KEY);
+        }
+
         //TODO: move all loading to worker thread
         loadSurvey(surveyId);
         loadLanguages();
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/akvo/flow/activity/FormActivity.java
```diff
@@ -560,35 +572,41 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
...
-            default:
-                mAdapter.onQuestionResultReceived(mRequestQuestionId, intent.getExtras());
-                break;
-        }
-
-        mRequestQuestionId = null;
+        // Set the result only after the QuestionView is loaded
+        mAdapter.setOnTabLoadedListener(new SurveyTabAdapter.OnTabLoadedListener() {
+            @Override
+            public void OnTabLoaded() {
```
valllllll2000:
```suggestion
            public void onTabLoaded() {
```

app/src/main/java/org/akvo/flow/ui/adapter/SurveyTabAdapter.java
```diff
@@ -47,6 +47,15 @@
     private List<QuestionGroup> mQuestionGroups;
     private List<QuestionGroupTab> mQuestionGroupTabs;
     private SubmitTab mSubmitTab;
+    private OnTabLoadedListener mOnTabLoadedListener;
+
+    public interface OnTabLoadedListener {
+        void OnTabLoaded();
```
valllllll2000:
```suggestion
        void onTabLoaded();
```

app/src/main/java/org/akvo/flow/ui/adapter/SurveyTabAdapter.java
```diff
@@ -88,6 +97,10 @@ private void loadTab(int position) {
             tab.load();
             tab.loadState();
             setupDependencies();// Dependencies might occur across tabs
+
+            if (mOnTabLoadedListener != null) {
+                mOnTabLoadedListener.OnTabLoaded();
```
valllllll2000:
```suggestion
                mOnTabLoadedListener.onTabLoaded();
```

nhternup:
Good catch. I guess one way would be to also track if the activity was restored and only then wait for the tab to load. Otherwise just call the onQuestionResultReceived directly as before.

PR #1410, opened 2019-06-18
Issue/1403 corrupt zip tests

valllllll2000:
NA
#### The solution
* Small refactor of file related methods to facilitate testing
* Add first tests to FileDataRepository and FileDataSource
More tests will be added in next PR
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


data/src/test/java/org/akvo/flow/data/repository/FileDataRepositoryTest.kt
```diff
@@ -0,0 +1,183 @@
...
+        verify(mockFileDataSource, times(1))!!.moveZipFiles()
+        verify(mockFileDataSource, times(1))!!.moveMediaFiles()
+        observer.assertNoErrors()
+    }
+
+    @Test
+    fun moveFilesShouldFailIfMoveZipFilesFails() {
+        val exception = Exception("test")
+        `when`(mockFileDataSource!!.moveZipFiles()).thenReturn(Completable.error(exception))
+        `when`(mockFileDataSource!!.moveMediaFiles()).thenReturn(Completable.complete())
```
muloem: A little detail here but if the focus is on zip files, maybe this one doesnt need to be in there?

data/src/test/java/org/akvo/flow/data/repository/FileDataRepositoryTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+        verify(mockFileDataSource, times(1))!!.moveZipFiles()
+        verify(mockFileDataSource, times(1))!!.moveMediaFiles()
+        observer.assertError(exception)
+    }
+
+    @Test
+    fun moveFilesShouldFailIfMoveMediaFilesFails() {
+        val exception = Exception("test")
+        `when`(mockFileDataSource!!.moveZipFiles()).thenReturn(Completable.complete())
```
muloem: Same here, if the focus is on media files then maybe zip files dont need to be in here?

data/src/test/java/org/akvo/flow/data/repository/FileDataRepositoryTest.kt
valllllll2000: if I remove the above line the test will fail as this method is called

data/src/test/java/org/akvo/flow/data/repository/FileDataRepositoryTest.kt
valllllll2000: in this case you are right but on the other hand the test if for the whole class and one of it's methods is that one plus I do not think I will have the occasion to revisit those tests again

Issue #1411, opened 2019-06-20
Ensure new Data Point (Record) identifiers are 14 chars long 


valllllll2000:
Copied from https://github.com/akvo/akvo-flow/issues/3117

When a data point identifier/UID is generated by com.gallatinsystems.surveyal.domain.SurveyedLocale.generateBase32Uuid(), there is a risk of it being truncated, ending up as abcd-efgh-jkm (or even shorter). This occurs if the number has sufficient number of leading zeroes before being converted to base-32.

This change will fix the issue but we should find a way to:
1) Use the same code in app and backend
2) Find a better way to identify the data points (maybe use a library)
3) Clarify the requirements: human readable, easy to remember, can be written on a paper, unique...
4) Test that solution for possible duplicates

valllllll2000:
## Test plan
any regression testing will catch issues. No point in testing as the occurrence of shorted uuids was so small that it is unlikely you will reproduce it.

PR #1415, opened 2019-06-25
[#1411] fixes issue with shorter record uuids

valllllll2000:
some record uids are shorter because the original uuid started with 0 or 00
#### The solution
The same as https://github.com/akvo/akvo-flow/blob/develop/GAE/src/com/gallatinsystems/surveyal/domain/SurveyedLocale.java#L260
Added some tests to make sure it works
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 
 # Testing
 .calabash_settings
+.kotlintest
 
 app/bin
 app/gen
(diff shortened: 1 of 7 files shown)


valllllll2000:
The failed tests are ui tests. I ran it twice and both times a different test failed. They are not related to the changes of this PR

PR #1420, opened 2019-07-10
Issue/1345 map list actions part3

valllllll2000:
NA
#### The solution
* latest version of the reactive library (unified naming)
* select/unselect map list items
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


app/src/main/java/org/akvo/flow/presentation/datapoints/map/offline/list/OfflineAreasListPresenter.java
```diff
@@ -132,4 +169,22 @@ public void onError(Throwable e) {
                 });
         disposables.add(subscribeWith);
     }
+
+    public void selectRegion(long regionId) {
+        Map<String, Object> params = new HashMap<>(2);
+        params.put(SaveSelectedOfflineArea.AREA_ID_PARAM, regionId);
+        saveSelectedOfflineArea.execute(new DisposableCompletableObserver() {
+            @Override
+            public void onComplete() {
```
kymni: do we need the method if it's empty?

valllllll2000: when implementing an interface, one needs to implement all methods unfortunately

PR #1431, opened 2019-07-23
[#1421] first part of the offline module separation

valllllll2000:
NA
#### The solution
In case we need to switch the map provider from mapbox it is better to have the offline maps feature as a separate module.
Part 1: added the manifest, build.gradle and other classes to the offline module
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


offlinemaps/src/main/java/org/akvo/flow/offlinemaps/di/PerFeature.java
```diff
@@ -0,0 +1,30 @@
...
+package org.akvo.flow.offlinemaps.di;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import javax.inject.Scope;
+
+@Scope
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PerFeature {
```
kymni: Do we need this class if it has no method?

valllllll2000: yes, its needed it's an interface that defines a dagger scope http://frogermcs.github.io/dependency-injection-with-dagger-2-custom-scopes/

Issue #1447, opened 2019-08-01
Use offline maps when capturing geoshapes 


janagombitova:
With 2.8.0 we completed the initial work on offline maps. You can create and save an offline map and view data points, or a single data point on it. However, when collecting geoshapes, we will use the old Google maps and offline maps are not supported here. 

With this issue we will: 
* NOT change how geoshapes work today (that is a larger task for another time) 
* ensure that when collecting a geoshape offline you can see the offline map and not work with a grey screen as today 

# Status quo

![geoshape](https://user-images.githubusercontent.com/12456965/62281390-fc4e9280-b44d-11e9-9244-312104c06bcc.gif)

Today when creating a geoshape you start with two action buttons in the top panel: 
* map tile selector 
* geoshape type selector - you need to select the geoshape type, to be able to proceed. 

If you are offline, the map screen will remain blank. 

# Change
How can we make it possible to use an offline map when creating new geoshapes? 
How can we make it possible to use an offline map when viewing a created geoshapes? 

@Kiarii and @valllllll2000  can you take a look at how geoshapes work today, change nothing about them but figure out how to make use of offline maps, as we do with data points?

valllllll2000:
I know this seems like an easy task but technically I still have to rewrite everything that is done using google maps into mapbox code so it's almost rewriting all from 0. If the design stays exactly the same then probably @Kiarii  does not need to get involved?? This different from datapoints as the logic for datapoints stayed the same. Here all the logic is related to drawing on the maps using a different library it will need to be rewritten.

janagombitova:
I see. My concern is that we do not know enough to adjust the implementation and on the other hand we know enough to know that once we start there are many rabbit holes to dive into (enforcing to decide on the shape type on the survey level so not allowing to decide in the field if it will be a line, points or a shape; validating the shapes, not allowing to create more than one shape per answer, etc). So improving the geoshapes is a large task on its own but does not have the priority right now. 

I would rather we finish up the offline maps, get fully rid of all google maps in the app and then later on when we prioritise needs around agricultural projects we pick up the project to clean up and improve geoshapes.

Kiarii:
@janagombitova are we then parking the issue for now?

janagombitova:
@Kiarii not this issue but re-working geoshapes from the user perspective. So Valeria will just reimplement the logic with offline maps, but we will not change anything around geoshapes until another time.

Kiarii:
so for now we only need consider the following right?
[...]

PR #1477, opened 2019-09-10
[#1447] create geoshapes mapbox part 3

valllllll2000:
NA
#### The solution
part 3 of the solution https://github.com/akvo/akvo-flow-mobile/pull/1471
Added shape related classes
NA
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
+++ b/app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
@@ -51,7 +51,7 @@ public class AreaShape extends Shape {
     public void removeSelectedPoint() {
         super.removeSelectedPoint();
         List<ShapePoint> points = getPoints();
-        //There can never be 3, as soon as there are 3 we add the first item to close the area
+        //Remove the extra point we added to create a valid shape since at this point the shape ceases to be valid
         if (points.size() == 3) {
             points.remove(points.size() - 1);
         }
--- a/app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
+++ b/app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2019 Stichting Akvo (Akvo Foundation)
+ *
+ * This file is part of Akvo Flow.
+ *
+ * Akvo Flow is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Akvo Flow is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
```diff
@@ -0,0 +1,79 @@
...
+        public AreaShape[] newArray(int size) {
+            return new AreaShape[size];
+        }
+    };
+
+    @Override
+    public void removeSelectedPoint() {
+        super.removeSelectedPoint();
+        List<ShapePoint> points = getPoints();
+        //There can never be 3, as soon as there are 3 we add the first item to close the area
```
muloem: The comment is not very clear for me.  Meaning if there are 4 points and one is removed to leave 3, then we add which first item to close the area?

app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
```diff
@@ -0,0 +1,79 @@
...
+    public void addPoint(LatLng latLng) {
+        unSelectAllPoints();
+        ShapePoint shapePoint = createSelectedShapePoint(latLng, getFeatureId());
+        List<ShapePoint> points = getPoints();
+        int size = points.size();
+        if (size < 2) {
+            points.add(shapePoint);
+        } else if (size == 2) {
+            points.add(shapePoint);
+            points.add(points.get(0));
```
muloem: In fact like 73 & 74 do exactly the same thing but twice right?  Would it work if you didn't have the first point added twice?

app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
valllllll2000: We need to add an extra point to close the shape. We only add it after 3 points have been added, so after adding
* 1 or 2 points -> invalid shape, user is still adding points, no closing yet
* 3 points -> shape is valid, add the first point (position 0) to close the shape
* 4 points or more -> nothing else needs to be done since the area is already closed. Closed means all points are linked to 2 other points.

app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
valllllll2000: no it's not the same, we add by default to the end, so the first add, adds to the end the current point, the second line adds the first point of the shape to close the shape. If I do not add the first point it would not work, the shape would not be correct.

app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
valllllll2000: but I see now the comment does not make sense here, will modify it a bit

app/src/main/java/org/akvo/flow/presentation/geoshape/entities/AreaShape.java
muloem: ok. Understood! 👍

PR #1493, opened 2019-09-19
Issue/1452 beta deployment

valllllll2000:
beta deployments with fabric are broken (fabric will be discontinued)
#### The solution
adding beta deployments to using firebase
using `./gradlew assembleFlowRelease appDistributionUploadFlowRelease` the app beta is sent to firebase and testers added to the "testers" group will receive an email to tests the app.
This is an alternative to using the url: https://uat1.akvoflow.org/app2 to download builds via an app that can be installed on a testers devices see screenshots.
We can have multiple versions ready to be downloaded in case we need to test updates.
In this case user does not need to type the url in the browser but just open the app.
In the app we can see different build versions for the same version name in case there were bug fixes.
Once this is merged, I can add it to bitrise as a step to release branches builds
#### Reviewer Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


valllllll2000:
Closing untill firebase has an app to simplify app download as this is not better than what we have on uat1

Issue #1555, opened 2019-12-16
Repeatable groups get copied


valllllll2000:
Steps to reproduce:
Create 4 repetitions of a repeatable question group but leave some answer empty except for the first repetition
Go back and reopen the datapoint which is saved
You will see that the answers from the first repetition are copied into the following repetitions even when the answer is supposed to be missing.

Bug was already present in version 2.8.0.1

valllllll2000:
All fixed

PR #1556, opened 2019-12-16
[#1555] added a "fix" for the question group bug

valllllll2000:
If a repeated question was left unanswered and you left the datapoint and opened the saved datapoint again, the first repetition was fine but after the second one, the responses were filled on the empty questions using the ones for the first repetition.
#### The solution
* repeated group responses will no longer be duplicated on empty questions
* added comments to explain and a TODO
NA
#### Reviewer Checklist
* [x] Added an explanation about the work done
* [x] Connected the PR and the issue on Zenhub

### DIFF
--- a/app/src/main/java/org/akvo/flow/ui/view/QuestionGroupTab.java
+++ b/app/src/main/java/org/akvo/flow/ui/view/QuestionGroupTab.java
@@ -201,15 +201,31 @@ public class QuestionGroupTab extends ConstraintLayout
         }
     }
 
+    //TODO: this method should be refactored to find a better way to distinguish between
+    //repeatable and non repeatable groups
     private void displayResponses() {
         Map<String, QuestionResponse> responses = mSurveyListener.getResponses();
         for (QuestionView qv : mQuestionViews.values()) {
             String questionId = qv.getQuestion().getId();
+            /*
+             * This works for questions without repetitions in format 123456
+             * or the questions whose repetition is not 0 the the key is 123456|1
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/akvo/flow/ui/view/QuestionGroupTab.java
```diff
@@ -201,15 +201,31 @@ private void updateGroupIterationHeaders() {
         }
     }
 
+    //TODO: this method should be refactored to find a better way to distinguish between
```
muloem: Create a separate issue for this and then we can clear out this comment

valllllll2000:
```suggestion
```

app/src/main/java/org/akvo/flow/ui/view/QuestionGroupTab.java
```diff
@@ -201,15 +201,30 @@ private void updateGroupIterationHeaders() {
         }
     }
 
+    //repeatable and non repeatable groups
```
valllllll2000:
```suggestion
```

app/src/main/java/org/akvo/flow/ui/view/QuestionGroupTab.java
valllllll2000: Done https://github.com/akvo/akvo-flow-mobile/issues/1557

Issue #1567, opened 2020-01-15
Improve display submitted form instance


valllllll2000:
Once the answer is submitted, it cannot be modified so we should display it separately from the editable form.
It should get all the answers for that instance and display them.
ViewPager will contain a collection of groups.
Each group is a fragment.
Fragments will handle their own data loading.
RecyclerView to display the scrollable content (the questions and answers)

PR #1585, opened 2020-02-13
[#1567] added ui classes for viewing form instance

valllllll2000:
#### The solution
This the first part of many others to display form which have already been exported (view only, the easier part)
Adding a few basic UI classes
Rename some constants to start on the right foot with our naming (Try using Form/Survey not SurveyGroup)
NA
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/presentation/form/view/FormViewActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+        val viewComponent =
+            DaggerViewComponent.builder()
+                .applicationComponent((application as FlowApp).applicationComponent)
+                .build()
+        viewComponent.inject(this)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        val surveyGroup = intent.getSerializableExtra(ConstantUtil.SURVEY_EXTRA) as SurveyGroup
```
valllllll2000: leaving this one for now as renaming the class would make the change in too many places

app/src/main/java/org/akvo/flow/presentation/form/view/FormViewActivity.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    override fun onCreateOptionsMenu(menu: Menu): Boolean {
+        menuInflater.inflate(R.menu.view_form_activity, menu)
+        return super.onCreateOptionsMenu(menu)
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.edit_lang -> {
+                //displayLanguagesDialog()
```
iperdomo: Is this going to be implemented?

valllllll2000: it will in future PRs

valllllll2000:
@iperdomo in the end not sure if you approved it or not, please confirm

gif

iperdomo:
@valllllll2000 it is approved, I use the wrong button when I commented one line

PR #1586, opened 2020-02-17
[#1567] added languages dialogs to form view

valllllll2000:
#### The solution
This the second part of many others to display form which have already been exported (view only, the easier part)
Adding classes related to form/survey languages selection and display inside the form
Rename some constants to start on the right foot with our naming (Try using Form/Survey not SurveyGroup)
NA
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/presentation/form/view/FormViewPresenter.kt
```diff
@@ -39,4 +41,21 @@ class FormViewPresenter @Inject constructor(): Presenter {
     ) {
         //TODO("not implemented")
     }
+
+    //language is saved per survey and not form???
```
tangrammer: is this an open question?

valllllll2000: it's a fact, I will remove the question mark

PR #1589, opened 2020-02-19
Issue/1567 display submitter form instance part3

valllllll2000:
#### The solution
3rd part of the refactor.
Added form languages save/update/view logic (from inside the menu)
NA
#### Reviewer Checklist


domain/src/main/java/org/akvo/flow/domain/repository/LanguagesRepository.kt
```diff
@@ -17,8 +17,12 @@
  * along with Akvo Flow.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package org.akvo.flow.presentation.form.languages
+package org.akvo.flow.domain.repository
```
valllllll2000: This is not a renamed class as github suggest, it's a new class

Issue #1611, opened 2020-04-07
Form download does not start correctly on some devices


valllllll2000:
**Describe the bug**
On some devices, when installing the app for the first time, the forms are not downloaded correctly

**To Reproduce**
1. Download new version of the app
2. Accept the walkthrough, login and accept the permissions
3. Forms are not downloaded, notification says 0 forms downloaded

**Expected behavior**
If there is an assignment, forms should be downloaded correctly without user having to take any extra steps.

**Device (please complete the following information):**
 - Device model: Moto g5
 - OS: Android 8.1.0
 - App Version 2.9.0.1

PR #1612, opened 2020-04-07
[#1611] fix issue with forms not being downloaded

valllllll2000:
When first downloading the app, the forms are not immediately downloaded and user has to manually restart the app
#### The solution
Fix the issue where the phone data was not properly updated once permissions were granted
NA
#### Reviewer Checklist


valllllll2000:
No, there is no cleanup, look closer https://github.com/akvo/akvo-flow-mobile/pull/1612/files#diff-b1973a5d304df63257793b1125aae7ebL50 deviceHelper is already there but we initialize the actual values in the constructor. The problem: the class is injected which means the dependency is created when the app starts. At this point we do not have the phone permissions yet so the imei, phone number.... etc are empty. When the user accepted those permissions it did not have an effect on the already initialized values and the assignments were not being correctly found for the device.

PR #1709, opened 2020-08-19
[#1567] more layout related classes

valllllll2000:
NA
#### The solution
This is a small part of a larger refactor of the forms display
Here we are adding the layouts of some question view types
There are some TODOs as this is not the final solution
NA
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/util/image/GlideImageLoader.java
```diff
@@ -112,9 +112,34 @@ public void loadFromFile(File file, ImageView imageView) {
...
+                            listener.onLoadFailed();
+                            return false;
+                        }
+
+                        @Override
+                        public boolean onResourceReady(Drawable resource, Object model,
+                                Target<Drawable> target, DataSource dataSource,
+                                boolean isFirstResource) {
+                            listener.onLoadSucceeded();
+                            return false;
```
tangrammer, line 133: do we need to return false in `onResourceReady` ?

valllllll2000: it's set to false because we update the image view ourselves because it needs some customization and not the library.

tangrammer (approved):

Issue #1717, opened 2020-09-21
Crash reported by Firebase related to permissions


valllllll2000:
**Crash**
When no permission is granted, the array is empty so we get "index out of bounds exception". This particular crash in when trying to take a picture while not having the permission to access storage granted. 

**Device :**
 - Device model: Samsung Galaxy A10s
 - OS: Android 9
 - App Version 2.9.4.1

PR #1718, opened 2020-09-21
[#1717] make sure permissions list is not empty to avoid crash

valllllll2000:
Crash in some cases when permissions array returns empty
#### The solution
Check if array is empty
NA
#### Reviewer Checklist

### DIFF
--- a/app/src/main/java/org/akvo/flow/event/PermissionAwareLocationListener.java
+++ b/app/src/main/java/org/akvo/flow/event/PermissionAwareLocationListener.java
@@ -58,7 +58,7 @@ public class PermissionAwareLocationListener {
     }
 
     public void handlePermissionResult(int requestCode, String[] permissions, int[] grantResults) {
-        if (grantResults == null || grantResults.length == 0) {
+        if (grantResults == null || grantResults.length == 0 || permissions.length == 0) {
             permissionListener.onPermissionNotGranted();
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/akvo/flow/event/PermissionAwareLocationListener.java
```diff
@@ -58,7 +58,7 @@ public void stopLocation() {
     }
 
     public void handlePermissionResult(int requestCode, String[] permissions, int[] grantResults) {
-        if (grantResults == null || grantResults.length == 0) {
+        if (grantResults == null || grantResults.length == 0 || permissions.length == 0) {
```
iperdomo, line 61: * In the first part of the `if` expression, we're considering the case that `grantResults` could be `null`. Could it be the same for `permissions`? 
* In other classes we're annotating parameters with `@Nullable` or `@NonNull`, can we do the same here?

iperdomo (changes requested):
Some minor comments on how to deal with `null` parameter values

app/src/main/java/org/akvo/flow/event/PermissionAwareLocationListener.java
valllllll2000: good find, I have fixed it

iperdomo (approved):

PR #1742, opened 2020-10-19
Issue/1733 datapoint download improvements

valllllll2000:
Datapoint download is quite slow especially when there are lots of images
If datapoint download stops halfway, when restarted all progress is lots
#### The solution
Use async for image download to download them in paralell
Only download images for assigned forms, sometimes a registration form is not assigned but we still get its data because the app needs them but since the data will not be displayed no need to download the image
Save cursor after each api call
NA
#### Reviewer Checklist


data/src/main/java/org/akvo/flow/data/entity/form/FormMapper.kt
```diff
@@ -25,6 +25,42 @@ import javax.inject.Inject
...
+            do {
+                val id = getIntColumnValue(cursor, SurveyColumns._ID)
+                val formId = getStringColumnValue(cursor, SurveyColumns.SURVEY_ID)
+                val surveyId = getIntColumnValue(cursor, SurveyColumns.SURVEY_GROUP_ID)
+                val formVersion = getStringColumnValue(cursor, SurveyColumns.VERSION)
+                val name = getStringColumnValue(cursor, SurveyColumns.NAME)
+                val type = getStringColumnValue(cursor, SurveyColumns.NAME)
+                val location = getStringColumnValue(cursor, SurveyColumns.NAME)
+                val filename = getStringColumnValue(cursor, SurveyColumns.NAME)
+                val language = getStringColumnValue(cursor, SurveyColumns.NAME)
```
muloem, line 40: I am  not sure about inner workings but here its ok for the second parameter to be `SurveyColumns.NAME`?

data/src/main/java/org/akvo/flow/data/entity/form/FormMapper.kt
```diff
@@ -25,6 +25,42 @@ import javax.inject.Inject
 
 class FormMapper @Inject constructor() {
 
+    fun mapForms(cursor: Cursor?): List<DataForm> {
+        val forms = mutableListOf<DataForm>()
+        if (cursor != null && cursor.moveToFirst()) {
+            do {
+                val id = getIntColumnValue(cursor, SurveyColumns._ID)
+                val formId = getStringColumnValue(cursor, SurveyColumns.SURVEY_ID)
+                val surveyId = getIntColumnValue(cursor, SurveyColumns.SURVEY_GROUP_ID)
```
muloem, line 34: Also wondering why the difference  i.e this is a `getInt` and the formId is a `getString`

valllllll2000: because one of them is saved as int value and the other as string in the database

data/src/main/java/org/akvo/flow/data/entity/form/FormMapper.kt
valllllll2000: oh no, that was an error I had copied from another part, will fix it

muloem (approved):

PR #1811, opened 2021-04-19
Issue/1810 bootstrap verification

valllllll2000:
NA
#### The solution
Use alias to check if the bootstrap survey is from the correct instance
Update tests
Converted some files to Kotlin
NA
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/service/bootstrap/BootstrapProcessor.kt
```diff
@@ -85,15 +85,15 @@ class BootstrapProcessor @Inject constructor(
             // now read the survey XML back into memory to see if there is a version
             val surveyMetadata = fileProcessor.readBasicSurveyData(surveyFile)
 
-            //most instances use the same for server base and instance url now
-            val instanceCodeName: String = if (BuildConfig.SERVER_BASE == BuildConfig.INSTANCE_URL) {
-                BuildConfig.AWS_BUCKET
+            if (surveyMetadata.alias.isNotEmpty() ) {
+                if (!BuildConfig.INSTANCE_URL.contains(surveyMetadata.alias)) {
```
muloem, line 89: Can this be passed incase there are two dashboards that contain similar alias even though the full string is not the same? e.g. `dashboard12` and `dashboard1`?

app/src/main/java/org/akvo/flow/service/bootstrap/BootstrapProcessor.kt
```diff
@@ -85,15 +85,15 @@ class BootstrapProcessor @Inject constructor(
...
+                    return ProcessingResult.ProcessingErrorWrongDashboard
+                }
+            } else if (surveyMetadata.app.isNotEmpty()) {
+                if (!BuildConfig.AWS_BUCKET.contains(surveyMetadata.app)) {
```
muloem, line 93: I am starting to wonder if its not better add a bucket attribute in the XML. That way even if we have the same bucket its explicit that one is the bucket and the other is the appId.

app/src/main/java/org/akvo/flow/service/bootstrap/BootstrapProcessor.kt
valllllll2000: no, for example, for uat1, INSANCE_URL="https://uat1.akvoflow.org" and alias would be "uat1.akvoflow.org", this would not work with uat2

app/src/main/java/org/akvo/flow/service/bootstrap/BootstrapProcessor.kt
valllllll2000: I am wondering... would it be useful? because as soon as these changes are added to the backend, the new alias will be added for all newly published forms. The old forms will keep having app attribute only, and that is why we still keep this check. In this scenario, I don't see the value of bucket, we would just use alias.

muloem (approved):

PR #1844, opened 2021-07-29
Issue/1813 support part2

valllllll2000:
#### The solution
part2 : adding a way to send 1 support ticket update with device information for now the id is set to default constant, complete solution see https://github.com/akvo/akvo-flow-mobile/pull/1845
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/presentation/settings/PreferencePresenter.kt
```diff
@@ -183,6 +198,46 @@ class PreferencePresenter @Inject constructor(
...
+                        "Android Version: $version" + "\r\n" +
+                        "Device model: $model" + "\r\n" +
+                        "App version: ${BuildConfig.VERSION_NAME}" + "\r\n" +
+                        "User name: $userName" + "\r\n" +
+                        "Device Identifier: $deviceId" + "\r\n" +
+                        "Instance: $instance"
+
+
+            val params = HashMap<String, Any>(4)
+            //val resId = 38964
```
valllllll2000, line 229: this will be fixed in part3

support/src/main/java/org/akvo/flow/support/net/ReamazeService.kt
```diff
@@ -0,0 +1,39 @@
...
+
+interface ReamazeService {
+
+    @Headers("Accept: application/json")
+    @POST("api/v1/conversations/{slug}/messages")
+    suspend fun uploadResponse(@Path("slug") slug: String, @Body messageResponse: MessageResponse): ResponseBody
+
+    @Headers("Accept: application/json")
+    @GET("api/v1/conversations/")
+    suspend fun fetchConversations(@Query("page") page: Int): ConversationsResult
```
muloem, line 38: Does this fetch only the latest/active conversations?

valllllll2000: unarchived conversations are fetched but there is a paginatiion so it the first page does not contain what we want we fetch the next one

muloem (approved):

PR #1845, opened 2021-07-29
Issue/1813 support part3

valllllll2000:
#### The solution
final solution to send full device information to support.
How does it work?
In a support ticket we realize that we need device information from one or more devices. We inform the partner to request that info directly from the device. In order to send it they will need the ref id, see screenshot
Screenshot from 2021-07-30 10-31-27

The device user will go to settings and find "send device info to support" and press that option. This will open a dialog and he then can write the 4 digits reference id. 
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/presentation/settings/SendDeviceInfoDialog.kt
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright (C) 2018-2019 Stichting Akvo (Akvo Foundation)
```
muloem, line 2: Is it a new file?

support/src/main/java/org/akvo/flow/support/domain/SendConversation.kt
```diff
@@ -57,29 +55,27 @@ class SendConversation @Inject constructor(
...
+            while (page < pageCount) {
                 conversationsResult = restApi.fetchConversations(page)
                 conversations = conversationsResult.conversations
-                totalCount += conversationsResult.totalCount
-                countLoaded = conversations?.size ?: 0
                 conversation = findConversation(conversations, refId)
                 if (conversation != null) {
                     return conversation
                 }
+                page++
```
muloem, line 78: Here maybe you want to do something like a `do  { ... } while ()`?

muloem: to make it simpler to read..

app/src/main/java/org/akvo/flow/presentation/settings/SendDeviceInfoDialog.kt
valllllll2000: no, I had copied from another file and forgot to update the copyright

support/src/main/java/org/akvo/flow/support/domain/SendConversation.kt
valllllll2000: done!

support/src/main/java/org/akvo/flow/support/domain/SendConversation.kt
```diff
@@ -57,30 +55,27 @@ class SendConversation @Inject constructor(
...
         var conversations = conversationsResult.conversations
-        var totalCount = conversationsResult.totalCount
-        var countLoaded = conversations?.size ?: 0
+        val pageCount = conversationsResult.pageCount
 
-        val refId = parameters[PARAM_REF_ID] as Int
         var conversation = findConversation(conversations, refId)
         if (conversation != null) {
             return conversation
         } else {
```
muloem, line 68: Isnt it possible to have all this within the `do ... while()`?

valllllll2000: done!

muloem (approved):

PR #1854, opened 2021-08-04
Issue/1851 group table

valllllll2000:
Part 1 of the work for the skip logic
#### The solution
moved table files to another package
added question group table
#### Reviewer Checklist


database/src/main/java/org/akvo/flow/database/tables/QuestionGroupTable.kt
```diff
@@ -0,0 +1,47 @@
...
+        const val COLUMN_REPEATABLE = "repeatable"
+        const val COLUMN_FORM_ID = "form_id"
+        const val COLUMN_ORDER = "order"
+
+        private const val CREATE_TABLE = ("CREATE TABLE IF NOT EXISTS " + TABLE_NAME + " ("
+                + COLUMN_ID + " INTEGER PRIMARY KEY AUTOINCREMENT,"
+                + COLUMN_GROUP_ID + " INTEGER DEFAULT -1"
+                + COLUMN_HEADING + " TEXT NOT NULL,"
+                + COLUMN_REPEATABLE + " INTEGER DEFAULT 0,"
+                + COLUMN_FORM_ID + " TEXT NOT NULL,"
```
muloem, line 44: Should this also be an integer just like the group id? Or the other way i.e the group id a text? They are both similar numbers.

database/src/test/java/org/akvo/flow/database/DatabaseHelperTest.java
```diff
@@ -73,7 +80,7 @@ public void onUpgradeShouldUpgradeCorrectlyIfVersionResponseMigrate() {
     @Test
     public void onUpgradeShouldUpgradeCorrectlyIfVersionTransmissionMigrate() {
```
muloem, line 81: Do we want to keep these old tests? in that case maybe its an idea to create a new constructor that adds in the question group table rather than using the old one?

database/src/test/java/org/akvo/flow/database/upgrade/UpgraderFactoryTest.java
```diff
@@ -67,7 +67,7 @@ public void createUpgraderShouldCreateCorrectUpgraderWhenAssignmentIteration() {
         UpgraderVisitor upgrader = (UpgraderVisitor) upgraderFactory
                 .createUpgrader(DatabaseHelper.VER_DATA_POINT_ASSIGNMENTS_ITERATION, mockDbHelper, mockDb);
```
muloem, line 68: e.g ..? should that be `VER_GROUPS`?

database/src/main/java/org/akvo/flow/database/tables/QuestionGroupTable.kt
valllllll2000: form id is always string in the other tables

database/src/test/java/org/akvo/flow/database/DatabaseHelperTest.java
valllllll2000: these tests test what should be upgraded depending on which version we are upgrading from, ie from version 1 to 2 there is only one upgrade done but 1 to 5 there are 5 of them

muloem (approved):

PR #1862, opened 2021-08-11
Issue/1851 group table part5

valllllll2000:
#### The solution
Remove all readonly (read only forms are viewed in another activity)
Use the same methods to load form and responses as read only activity (getting groups from database)
Fix some issues after the change
#### Reviewer Checklist


app/src/main/java/org/akvo/flow/presentation/form/OldFormMapper.kt
```diff
@@ -0,0 +1,158 @@
...
+            val altText = domainAltText[language]
+            if (altText != null) {
+                textMap[language] = AltText(altText.languageCode, altText.type, altText.text)
+            }
+        }
+        return textMap
+    }
+
+    fun mapResponses(responses: List<Response>): HashMap<String, QuestionResponse> {
+        // QuestionId - QuestionResponse
```
valllllll2000, line 154: this will be added in part6

utils/src/main/java/org/akvo/flow/utils/XmlFormParser.kt
```diff
@@ -322,8 +323,227 @@ class XmlFormParser @Inject constructor(private val helper: FileHelper) {
...
+            parser.setFeature(XmlPullParser.FEATURE_PROCESS_NAMESPACES, false)
+            parser.setInput(inputStream, null)
+            var eventType = parser.eventType
+            while (eventType != XmlPullParser.END_DOCUMENT) {
+
+                when (eventType) {
+                    /**
+                     * Beginning of a tag for example <question>
+                     */
+                    XmlPullParser.START_TAG -> {
```
muloem, line 351: Just a thought based on an example of something I saw, wouldnt it be easier if the sections here were broken down into separate functions per type of element? e.g. whats done here..

https://developer.android.com/training/basics/network-ops/xml#parse

valllllll2000: it's not that easy because we have some sections that are common to questions, options , but I can take a look at the end of the work done of how it can be improved

muloem (approved):

PR #1863, opened 2021-08-13
Issue/1851 group table part6

valllllll2000:
#### The solution
Using the new loading in FormActivity
cleanup some fields
#### Reviewer Checklist


domain/src/main/java/org/akvo/flow/domain/entity/DomainResponse.kt
```diff
@@ -0,0 +1,29 @@
...
+package org.akvo.flow.domain.entity
+
+data class DomainResponse(
+    val id: Long,
+    val questionId: String,
+    val answerType: String,
+    val value: String,
+    val iteration: Int,
+    val isIncludeFlag: Boolean
+)
```
muloem, line 29: there is no `insanceId` field here?

valllllll2000: no, not needed, for now at least

muloem (approved):

PR #1866, opened 2021-08-16
Issue/1851 group table part7

valllllll2000:
#### The solution
Bootstrap uses the same code to insert surveys/forms
#### Reviewer Checklist


data/src/main/java/org/akvo/flow/data/repository/FormDataRepository.kt
```diff
@@ -94,6 +105,63 @@ class FormDataRepository @Inject constructor(
...
+            if (!instanceUrl.contains(surveyMetadata.alias)) {
+                throw WrongDashboardError()
+            }
+        } else if (surveyMetadata.app.isNotEmpty()) {
+            if (!awsBucket.contains(surveyMetadata.app)) {
+                throw WrongDashboardError()
+            }
+        } else {
+            throw WrongDashboardError()
+        }
```
muloem, line 148: I was wondering about this section. Shouldnt they be multiple `if` statements rather than `if.. else`.  Say for instance alias is not empty, and is correct but the app info is incorrect, that will be skipped.  I guess the assumption is that once one is correct then the other should be but for example the recent migrations of instances may invalidate that assumption.

valllllll2000: alias has priority over app, that is why we added it so in this case app does not matter

muloem (approved):

PR #1868, opened 2021-08-17
Issue/1851 group table part8

valllllll2000:
#### The solution
migrate database: insert all groups from all forms
#### Reviewer Checklist


database/src/main/java/org/akvo/flow/database/britedb/BriteSurveyDbAdapter.java
```diff
@@ -565,6 +565,9 @@ public void updateSurvey(ContentValues updatedValues, String surveyId) {
     public void deleteAllSurveys() {
         briteDatabase.delete(Tables.SURVEY, null);
         briteDatabase.delete(Tables.SURVEY_GROUP, null);
+        briteDatabase.delete(QuestionGroupTable.TABLE_NAME, null);
+        briteDatabase.delete(FormUpdateNotifiedTable.TABLE_NAME, null);
+        briteDatabase.delete(LanguageTable.TABLE_NAME, null);
```
muloem, line 570: Any reason not to follow the same naming convention as the previous tables. i.e. `Tables.SURVEY`?

database/src/main/java/org/akvo/flow/database/migration/GroupsMigrationHelper.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.io.FileInputStream
+import java.io.FileNotFoundException
+import java.util.ArrayList
+
+class GroupsMigrationHelper {
+
+    fun migrateGroups(db: SQLiteDatabase, context: Context) {
+        val cursor =
+            db.query(Tables.SURVEY, arrayOf(SurveyColumns._ID), null, null, null, null, null)
+        val formIds = ArrayList<String>()
```
muloem, line 41: Maybe this should be something like a `Set` to remove duplicates or is that already handled in the query?

database/src/main/java/org/akvo/flow/database/britedb/BriteSurveyDbAdapter.java
valllllll2000: the new tables I have been adding, have code inside their own class with table name and columns and create code. I find it more clear than having Tables just the list of tables and then columns somewhere else. Also the DatabaseHelper becomes really large with lots of code so I separated it this way

database/src/main/java/org/akvo/flow/database/migration/GroupsMigrationHelper.kt
valllllll2000: that's a good point

database/src/main/java/org/akvo/flow/database/migration/GroupsMigrationHelper.kt
```diff
@@ -0,0 +1,76 @@
...
+import java.io.FileNotFoundException
+
+class GroupsMigrationHelper {
+
+    fun migrateGroups(db: SQLiteDatabase, context: Context) {
+        val cursor =
+            db.query(Tables.SURVEY, arrayOf(SurveyColumns._ID), null, null, null, null, null)
+        val formIds = HashSet<String>()
+        if (cursor != null && cursor.moveToFirst()) {
+            val columnIndex = cursor.getColumnIndex(QuestionGroupTable.COLUMN_FORM_ID)
```
muloem, line 42: Should this be the form id column?

muloem (approved):

